PR #1, opened 2021-08-19
Add Tree and TreeUpdate unit tests

DataTriny:


mwcampbell:
Thanks!

I haven't really reviewed the test code yet, but I noticed that the `new_node` function explicitly specifies the default values for all of the fields other than `id` or `node`. It's safe to omit these and let Rust implicitly use the defaults. I'd rather do it that way.

mwcampbell:
The code should also omit default fields when initializing other structs, like `TreeUpdate` and the schema's `Tree`.

Don't forget to run `cargo fmt` and `cargo clippy`. When we add a CI, we'll want it to enforce those things.

DataTriny:
To allow omitting fields when initializing those structs, they would have to implement the `Default` trait. However, I am not sure if a default value is applicable to `Node.id` or `Tree.id` for instance. I admit the `new_node` is kinda ugly though.

mwcampbell:
You're right, `Node::id` and `Tree:id` should not have default values. Neither should `Node::role` IMO.

But it apparently isn't necessary for these structs to derive the `Default` trait. I tried modifying `new_node` to omit everything except `id` and `role`, and it does compile.

mwcampbell:
My mistake! I forgot that `cargo build` doesn't compile tests.

So we probably need to add a `Node::new` function in the schema crate.

DataTriny:
I would prefer adding `Node::new(NodeId, Role)` to `schema` of course.

How about `Tree:new(TreeId, root: NodeId)`? Or do we require the focused element, string encoding and so on?

mwcampbell:
Yes, `Node::new` should definitely have `id` and `role` parameters.

`Tree:new` should have `id` and `source_string_encoding` parameters IMO. Also, remember that the `root` field is on `TreeUpdate`, not `Tree`. I'm thinking about changing that though; I don't really see a good reason for the way that Chromium did that.

DataTriny:
@mwcampbell all changes applied.

mwcampbell:
A few more suggestions:

I think the test tree ID should be declared as a constant as well, since it's used several times.

The string encoding should be UTF-8. People will be looking at our tests for examples of how to use AccessKit, and when a tree is generated by Rust code, the source string encoding will naturally be UTF-8. The UTF-16 option is for trees that were generated in languages that natively use UTF-16, like JavaScript and JVM languages.

I think `add_cuild_to_root_node` should test that the child has the correct parent, like `root_node_has_children`.

Finally, while I'm still deciding on the conventions that AccessKit will use for tree structure, I think it's safe to assume that, at least on most platforms, the root node of a tree will have a role of `Windo`, not `Application`. The `Application` role in AccessKit, like its counterpart in Chromium, is for exposing the ARIA `application` role, not for being the parent of all windows as in AT-SPI. So I think you should modify all of the tests so the root node is a `Window`. For the focus test, that means you can eliminate a level from the hierarchy. For the other tests with children, that means that the children should be something other than windows, such as buttons.

PR #3, opened 2021-08-30
[WIP] Node methods to retrieve siblings and unignored children

DataTriny:
This took me longer than expected.

I am not really proud of the `UnignoredChildren` iterator implementation, I don't find it elegant but I can't think of another way to do it right now.

## Questions

- Should we implement `UnignoredChildren::len()`? I can't find a good value to return for `UnignoredChildren::size_hint` because you'll have to iterate to the end to get the correct size, so maybe the default is fine.
- I'm not sure about the `unignored_children_reverse` test first assertion. The root node clearly has more unignored children, but given that one descendant got an unignored parent, it's a stopping condition. I couldn't find such test case inside Chromium's codebase.

### DIFF
--- a/consumer/src/lib.rs
+++ b/consumer/src/lib.rs
@@ -8,4 +8,4 @@ pub(crate) mod tree;
 pub use tree::{Reader as TreeReader, Tree};
 
 pub(crate) mod node;
-pub use node::{Node, WeakNode};
+pub use node::{FollowingSiblings, Node, PrecedingSiblings, WeakNode};
--- a/consumer/src/node.rs
+++ b/consumer/src/node.rs
@@ -68,39 +68,24 @@ impl Node<'_> {
 
     pub fn unignored_children(
         &self,
+    ) -> impl DoubleEndedIterator<Item = Node<'_>> + FusedIterator<Item = Node<'_>> {
+        UnignoredChildren::new(self).map(move |id| self.tree_reader.node_by_id(id).unwrap())
+    }
+
+    pub fn following_siblings(
+        &self,
     ) -> impl DoubleEndedIterator<Item = Node<'_>>
            + ExactSizeIterator<Item = Node<'_>>
            + FusedIterator<Item = Node<'_>> {
-        self.data()
-            .children
-            .iter()
-            .map(move |id| self.tree_reader.node_by_id(*id).unwrap())
+        FollowingSiblings::new(self).map(move |id| self.tree_reader.node_by_id(id).unwrap())
     }
 
-    pub fn following_siblings(&self) -> FollowingSiblings {
-        FollowingSiblings {
-            delta: 1,
-            index_in_parent: self
-                .state
-                .parent_and_index
-                .as_ref()
-                .map_or(0, |ParentAndIndex(_, index)| *index),
-            parent: self.parent(),
-            reader: self.tree_reader,
-        }
-    }
-
-    pub fn preceding_siblings(&self) -> PrecedingSiblings {
-        PrecedingSiblings {
-            delta: 1,
-            index_in_parent: self
-                .state
-                .parent_and_index
-                .as_ref()
-                .map_or(0, |ParentAndIndex(_, index)| *index),
-            parent: self.parent(),
-            reader: self.tree_reader,
-        }
+    pub fn preceding_siblings(
+        &self,
+    ) -> impl DoubleEndedIterator<Item = Node<'_>>
+           + ExactSizeIterator<Item = Node<'_>>
+           + FusedIterator<Item = Node<'_>> {
+        PrecedingSiblings::new(self).map(move |id| self.tree_reader.node_by_id(id).unwrap())
     }
 
     pub fn global_id(&self) -> String {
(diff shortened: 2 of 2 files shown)


DataTriny:
Forget my second question: I wanted to test reversed unignored children of root, but fetched those from paragraph 3 instead.

Now I will try to separate `UnignoredChildren` and `UnignoredSiblings` because we might need those as well at some point.

DataTriny:
@mwcampbell There are still plenty of methods that will require some node iterators.

I'd like to have your opinion on what I have so far before I start implementing the remaining ones.

Thanks.

DataTriny:
Going again through the list of methods on `Node`, here are those missing from this PR that are related to children/siblings/parent:

- `GetUnignoredIndexInParent()`: it relies on a cached value initially set in the constructor, I can implement the algorithm but it will be slower if performed on each call.
- `*CrossingTreeBoundary()`: these variants of other methods involve nested trees, since we don't have such thing yet, I skipped them.
- `GetNextUnignoredInTreeOrder()` and `GetPreviousUnignoredInTreeOrder()`: I haven't been able to locate a single use of them. Could you confirm @mwcampbell please?
- `IsChildOfLeaf()`, `IsEmptyLeaf()`, `IsLeaf()`: these also involve subtrees.

Apart from that, I would like to have your opinion on the order of methods, I haven't thought too much about it when I added the new ones. Since `Node` will end up having a lot of methods, do we want to arrange them alphabetically, or in a way easier to read in the docs (probably with `id()`, `role()`, etc... at the top?

mwcampbell:
I think it's reasonable to defer implementing all of the methods you mentioned until we need them.

I think the method order should be logical, not alphabetical.

mwcampbell:
I've relicensed AccessKit as we discussed. Please replace the copyright header comment in iterators.rs.

DataTriny:
`iterators.rs` relicensed. PR should more or less be in a reviewable state by now.

mwcampbell:
I think the iterators for unignored nodes should return `Node` (the one in the consumer crate), not `NodeId`. These iterators already have to hold a `TreeReader` and instantiate `Node`s to determine which ones are unignored, so returning `NodeId`s and then requiring the caller to re-instantiate the `Node`s seems wasteful. But if there's a lifetime complication I've overlooked, then the current implementation is acceptable.

mwcampbell:
The constructors for `FollowingSiblings` and `PrecedingSiblings` expose a weakness in my `Node` API. Because `Node` has separate `parent` and `index_in_parent` functions, you have to do two independent checks of optional values. But we know that the parent and index are both either there or not, so having to check them independently is suboptimal. I suggest replacing `Node::index_in_parent` with a `parent_and_index` function, which would return the `parent_and_index` field of the underlying data. This implies that the `ParentAndIndex` struct would need to be public. I'm open to other ideas.

mwcampbell:
I was mistaken; the existing `ParentAndIndex` struct wouldn't need to be made public, because the return value of the `Node::parent_and_index` function would contain a `Node` for the parent, not a `NodeId`.

mwcampbell:
I wonder if we actually need custom iterators for `following_siblings` and `preceding_siblings`. Would it be possible to take a slice of the parent's `children` field, make an iterator over that slice, then map over that? I ask because the `FollowingSiblings` and `PrecedingSiblings` iterators are a lot like the generic slice iterator, but probably not as well optimized.

DataTriny:
Addressing your points:

- Unfortunately I haven't been able to make `UnignoredChildren` yield `Node` because `Node` need a lifetime parameter, which conflicts with `&self`. This worries me because all those extra `HashMap` lookups seem very inefficient to me, if you think of a better way I would be glad to know it.
- Not sure I understood you on the `Node::parent_and_index`: yes having to check both parent and index individually is annoying, but `ParentAndIndex` only contain parent's `NodeId`, so we would need to fetch and unwrap it anyway. It's a slight improvement, but maybe returning `(Node<'_>, usize)` would even be better?
- I first tried to use already existing iterators for `(Following/Preceding)Siblings` but couldn't find a combination of methods to express the fact that if we don't have a parent, then we should return an empty iterator. Furthermore, I found it clearer to separate the preceding and following siblings.

mwcampbell:
To respond to your previous responses:

* If having the iterators yield `Node` introduces lifetime-related challenges, then we'll stick with hash table lookups for now.
* I had in mind a `Node::parent_and_index` function that returns `Option(<Node<'_>, usize)>`. It should return `None` if either there's no parent or the attempt to look up the parent fails. So we still have to do a double check, but only in that one place.
* I agree that having separate functions for returning the preceding and following siblings makes sense. But instead of using custom iterators, these functions could conditionally either use `std::iter::empty`, or `map` over a standard slice iterator. Am I missing something?

mwcampbell:
Ah, I think I see the problem with conditionally returning either `std::iter::empty` or a map over a slice iterator. Those two branches have different return types, so we'd have to take the performance hit of boxing and dynamic dispatch. Your custom iterators are more efficient in the common case, so let's stick with those.

DataTriny:
Sorry, I of course meant having `Node::parent_and_index` return an optional tuple, not just a tuple. I'll implement it, and will do some more research on iterators that yield structs with lifetimes, but I'm afraid we'll have to stick with my first approach...

I noticed that you kept the comments from Chromium while porting, which I didn't do at all. So I will add them too.

mwcampbell:
I'd like to get this merged today if we can, so I can start using it in my new windows-basics branch.

It would be fine with me to add the `Node::parent_and_index` method in a future PR.

If `consume/src/iterators.rs` has substantial code adapted from Chromium, then it should include the secondary copyright header, like the one I put in `schema/src/lib.rs`.

DataTriny:
I will be available in ~3 hours to work on this, and it should be very quick.

DataTriny:
@mwcampbell everything mentionned here has been addressed.

DataTriny:
No idea how it happened...

mwcampbell:
One more little thing: I actually think it's OK for the root of this particular test tree to have a role of `RootWebArea`. The tree is clearly meant to have the structure of a web page, and when AccessKit eventually supports that level of complexity, we'll probably recommend that web pages are encapsulated in subtrees, as they are in Chromium. Having a role of `Window` on the root node isn't an absolute rule for AccessKit, though it's probably what we'll do for the simple UIs that AccessKit will practically support at first.

Other than that, it looks good. If you don't mind, please change that role back to `RootWebArea`. Then I'll merge.

mwcampbell:
I'm sorry, I have one other piece of feedback: Why does `is_descendant` take an `Option<&Node>`? I guess it makes the tests a littler simpler, as you can directly pass the result of `node_by_id`. But is there any real, non-test, use-case where one would want to pass `None` to this function? If not, I think it's important to get the API right at the slight expense of the tests.

DataTriny:
Actually I changed that root element recently and I forgot to ask you if we would be OK having a `RootWebArea`.

I modified `is_ancestor_of`: I initially put an `Option` because the original method accepts null values as input, but they explicitly check for null values before calling it (ui/accessibility/platform/ax_platform_node_base.cc) so I think it's unnecessary.

mwcampbell:
Oops, looks like I ended up duplicating the PR title in the commit message. Now I know.

Anyway, thanks @DataTriny for all of your work on this.

PR #40, opened 2021-10-03
Refactor consumer Node and iterators to reduce lookups

### LINKED ISSUE #4: Need methods for traversing nodes, including unignored nodes
Belatedly opening an issue for this. #3 is addressing this; I just need to review it.

mwcampbell:
The key insight that made this possible was the realization that `Node` could derive the `Copy` trait, which meant that `Node` methods that return other `Node`s don't need to borrow `self`.

Fixes #4

### DIFF
--- a/consumer/src/iterators.rs
+++ b/consumer/src/iterators.rs
@@ -13,7 +13,6 @@ use std::iter::FusedIterator;
 use accesskit_schema::NodeId;
 
 use crate::node::Node;
-use crate::tree::Reader as TreeReader;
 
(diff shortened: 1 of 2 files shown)


mwcampbell:
@DataTriny What do you think of this?

DataTriny:
I feel kinda stupid right now: I knew that deriving `Copy` would help with these iterators but for some reason I thought this would not be possible due to the fact that `NodeData` contains `String`s, but since we only store a reference to a `NodeState` then indeed it works! Nice catch!

The code looks good to me, you can proceed to merge.

PR #41, opened 2021-10-09
Use boxed `str` and slices to reduce size of `accesskit_schema::Node`

mwcampbell:
This reduces the size of `Node` by 296 bytes (on x64), from 1568 bytes down to 1272 bytes, with no loss of functionality, since the contents of `Node` aren't meant to be mutated in-place.


DataTriny:
That's clever indeed. I approve!

PR #42, opened 2021-10-11
Support offset containers in `accesskit_consumer::Node::bounds`

mwcampbell:


DataTriny:
Well... I was actually working on this exact same feature.

I would also have tested a non-zero top/left `Rect` with `offset_container != None`.

Do you think `Rect`' should be able to exceed `offset_container`'s bounds?

Should we ensure positive `Rect` `width/height`?

PR #44, opened 2021-10-12
Box `NodeState` to save memory

mwcampbell:
The `accesskit_schema::Node` struct is fat enough that storing instances of it inline in a growable collection wastes significant memory. That's because, as the collection grows, it allocates more than it needs at the moment, in anticipation of more growth. So, to save memory, the `HashMap` in `accesskit_consumer::Tree` now puts the `NodeState` values inside a `Box`. In my testing on Linux/x64, this saves about 550 KB for a fairly small (267-node) tree converted from Chromium, and the overhead of dereferencing the extra pointer is insignificant.

### DIFF
--- a/consumer/src/tree.rs
+++ b/consumer/src/tree.rs
@@ -17,7 +17,7 @@ pub(crate) struct NodeState {
 }
 
 pub(crate) struct State {
-    pub(crate) nodes: HashMap<NodeId, NodeState>,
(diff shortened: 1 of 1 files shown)


DataTriny:
Code is OK.

We really have to do something about this struct, it's way to big. Maybe start by using bitflags for all the boolean states it contain?

PR #45, opened 2021-10-12
Use `parking_lot` crate for RW locks

mwcampbell:
This speeds up acquisition of an uncontended RW lock for reading (the common case). On Linux/x64, using a release build with LTO, the time to upgrade a `WeakNode` in my benchmark is reduced by about 30%. The improvement is less dramatic on Windows, but still an improvement.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,6 +7,7 @@ name = "accesskit_consumer"
 version = "0.1.0"
 dependencies = [
  "accesskit_schema",
+ "parking_lot",
 ]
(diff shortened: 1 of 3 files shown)


mwcampbell:
Ah, missed the bit about wasm.

I still think it's worth doing this. Another nice thing about `parking_lot` is the `downgrade` method on `RwLockWriteGuard`. I think I'm going to use that when collecting events after updating a tree. So I'll go ahead and merge.

PR #46, opened 2021-10-12
Add basic tracking of the changes in a tree update

mwcampbell:
The code for processing a `TreeUpdate` can now keep track of the nodes that are added, updated, or removed, as well as a change in focus. This feature will be the basis for platform-specific events.

### DIFF
--- a/consumer/src/tree.rs
+++ b/consumer/src/tree.rs
@@ -504,7 +504,11 @@ mod tests {
         };
         let mut got_focus_change = false;
         tree.update_and_process_changes(second_update, |change| {
-            if let super::Change::FocusMoved { old_id, new_node: Some(new_node) } = &change {
+            if let super::Change::FocusMoved {
+                old_id,
+                new_node: Some(new_node),
+            } = &change
+            {
                 if *old_id == Some(NODE_ID_2) && new_node.id() == NODE_ID_3 {
                     got_focus_change = true;
                     return;
--- a/consumer/src/tree.rs
+++ b/consumer/src/tree.rs
@@ -26,8 +26,13 @@ pub(crate) struct State {
 
 enum InternalChange {
     NodeAdded(NodeId),
-    NodeUpdated(Box<NodeData>),
-    FocusMoved((Option<NodeId>, Option<NodeId>)),
+    NodeUpdated {
+        old_data: Box<NodeData>,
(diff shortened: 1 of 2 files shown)


DataTriny (commented):
Overall you are on the right track, just some comments on naming and tests.

mwcampbell:
I changed the names of the variants, and switched to named fields where appropriate.

I agree that the tests are awkward. But in the current design, I can't collect the changes in a `Vec` because some of the variants implicitly borrow a `tree::Reader` which is only alive during the `update_and_process_changes` function. That's why I decided to use a callback in the first place. I'm open to something better than `assert!(false)` though.

consumer/src/tree.rs
```diff
@@ -301,7 +392,28 @@ mod tests {
...
+                    return;
+                }
+            }
+            if let super::Change::NodeAdded(node) = &change {
+                if node.id() == NODE_ID_2 {
+                    got_new_child_node = true;
+                    return;
+                }
+            }
+            assert!(false);
```
DataTriny, line 413: You could directly `panic!` with a message such as: Expected root node update and new child changes. At least it will be clear from the test results if it ever break.

consumer/src/tree.rs
```diff
@@ -334,7 +446,28 @@ mod tests {
...
+                    return;
+                }
+            }
+            if let super::Change::NodeRemoved(old_data) = &change {
+                if old_data.id == NODE_ID_2 {
+                    got_removed_child_node = true;
+                    return;
+                }
+            }
+            assert!(false);
```
DataTriny, line 467: `panic!("Expected root node update and child removal changes.");`

consumer/src/tree.rs
```diff
@@ -369,8 +502,99 @@ mod tests {
...
+        tree.update_and_process_changes(second_update, |change| {
+            if let super::Change::FocusMoved { old_id, new_node } = &change {
+                if let Some(new_node) = new_node {
+                    if *old_id == Some(NODE_ID_2) && new_node.id() == NODE_ID_3 {
+                        got_focus_change = true;
+                        return;
+                    }
+                }
+            }
+            assert!(false);
```
DataTriny, line 515: `panic!("Expected focus moved change.");`

PR #49, opened 2021-10-19
Basic UIA implementation for Windows

mwcampbell:
The UIA implementation has the following features so far:

* Tree navigation
* Control type and name properties
* Focus tracking
* Property change events for all supported properties

This PR also includes an example program that demonstrates all of the above.

I think I'll add automated tests in a future PR.


platforms/windows/src/tests/simple.rs
```diff
@@ -0,0 +1,180 @@
...
+
+use super::*;
+
+const WINDOW_TITLE: &str = "Simple test";
+
+const WINDOW_ID: NodeId = NodeId(unsafe { NonZeroU64::new_unchecked(1) });
+const BUTTON_1_ID: NodeId = NodeId(unsafe { NonZeroU64::new_unchecked(2) });
+const BUTTON_2_ID: NodeId = NodeId(unsafe { NonZeroU64::new_unchecked(3) });
+
+fn get_button_1(name: &str) -> Node {
```
DataTriny, line 19: Same comment as in the example: do we need two functions for this?

schema/src/lib.rs
```diff
@@ -502,10 +502,12 @@ pub enum StringEncoding {
     Utf16,
 }
 
-/// The stable identity of a node, unique within the node's tree.
 // This is NonZeroU64 because we regularly store Option<NodeId>.
+pub type NodeIdContent = std::num::NonZeroU64;
+
+/// The stable identity of a node, unique within the node's tree.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, Deserialize, JsonSchema)]
-pub struct NodeId(pub std::num::NonZeroU64);
```
DataTriny, line 508: Is this change necessary? `NonZeroU64` and `NodeId` should take the same amount of memory and a tuple struct does not modify the memory layout of the underlying type, right?

DataTriny (commented):
Very happy with all of this overall. I think we are close to merging this!

DataTriny:
Two general remarks:

- How would the example application handle old Win32 controls that implement MSAA? There are plenty of GUI toolkits that rely on native Windows menu bars, so that is something to check.
- This PR introduces a bunch of TODOs (and that is fine at this stage). But, do you keep a list? Opening issues for each one would help to keep track of remaining work.

mwcampbell:
@DataTriny To answer your question about toolkits that use MSAA-based Win32 controls: First, for the specific case of a Win32 menu bar, UIA will automatically take care of that, as it does for the title bar; we would have to specifically turn that off through the provider options. UIA will also automatically add child windows to the tree. UIA has complex logic for combining multiple types of providers into one tree. I saw that logic while I was on the Windows accessibility team at Microsoft, though I've forgotten a lot of the details in the past year since I left.

PR #52, opened 2021-10-29
Raise node update events when focus changes

mwcampbell:
The cross-platform change tracking code now raises `NodeUpdated` events when the focus moves, but only if the node in question wasn't already updated as part of the tree update. This allows platform adapters to look for platform-specific property changes on the previously and newly focused nodes, again without duplicate events.


DataTriny:
Approved!

PR #55, opened 2021-11-03
Make serde and schemars optional

mwcampbell:
This reduces build times when these features aren't needed.

### DIFF
--- a/schema/Cargo.toml
+++ b/schema/Cargo.toml
@@ -5,5 +5,5 @@ authors = ["Matt Campbell <mattcampbell@pobox.com>"]
 edition = "2018"
 
 [dependencies]
-schemars = "0.8.3"
-serde = { version = "1.0", features = ["derive"] }
+schemars = { version = "0.8.3", optional = true }
+serde = { version = "1.0", features = ["derive"], optional = true }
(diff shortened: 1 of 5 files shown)


schema/Cargo.toml
```diff
@@ -5,5 +5,5 @@ authors = ["Matt Campbell <mattcampbell@pobox.com>"]
 edition = "2018"
 
 [dependencies]
-schemars = "0.8.3"
-serde = { version = "1.0", features = ["derive"] }
+schemars = { version = "0.8.3", optional = true }
+serde = { version = "1.0", features = ["derive"], optional = true }
```
DataTriny, line 9: We'll require serde to send our types through D-Bus. So any user of AccessKit who want to support Linux will need serde. I don't know if it really make sense to make it optional then.

mwcampbell: Sure, the AT-SPi adapter crate will need to have a required dependency on serde. But that doesn't mean we need to force the use of serde for the AccessKit schema types. So this change will still reduce the build time for the schema crate, even on Linux, and of course, it will eliminate the dependency on serde for Windows and Mac users of Rust-based toolkits (where serialization of schema types isn't needed).

DataTriny (approved):

PR #58, opened 2021-11-05
Windows: Handle focus corner cases in example and tests

mwcampbell:
To see these cases in action, run the example, open the system menu (Alt+Space), choose Move or Size, then exit the move/size mode with Escape. Or open the System menu, escape out of that menu, then escape out of menu mode. In both cases, focus should not return to the button until you leave the special mode.


DataTriny (approved):
I never use the keyboard to perform such tasks and, to be honest, all the applications I could find (including Windows ones) seem to behave incorrectly.

PR #59, opened 2021-11-05
Windows: More work on basic UIA properties and navigation

mwcampbell:
While testing with JAWS, I discovered that I need to implement the `IsControlElement`, `IsContentElement`, and `IsEnabled` properties, because JAWS doesn't respect the normal UIA-provided default values for these properties. I also refactored the abstraction for UIA properties so the getters don't all have to return an `Option`. And I modified navigation to traverse all elements, since we'll let UIA filter elements via `IsControlElement` and `IsContentElement`.

### DIFF
--- a/platforms/windows/src/util.rs
+++ b/platforms/windows/src/util.rs
@@ -4,7 +4,7 @@
 // the LICENSE-MIT file), at your option.
 
 use std::{
-    convert::{From, TryInto},
+    convert::{From, Into, TryInto},
     mem::ManuallyDrop,
 };
 use windows::Win32::{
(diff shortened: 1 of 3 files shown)


platforms/windows/src/util.rs
```diff
@@ -4,7 +4,7 @@
 // the LICENSE-MIT file), at your option.
 
 use std::{
-    convert::{From, TryInto},
+    convert::{From, Into, TryInto},
```
DataTriny, line 7: `From` and `Into` are already in the prelude, no need to import them here.

```suggestion
    convert::TryInto,
```

DataTriny (changes requested):
Just a detail.

PR #62, opened 2021-11-16
Safer mutation of WindowState in Windows tests

mwcampbell:
After seeing how druid-shell handles window state, I decided it would be safer to use a const pointer plus interior mutability, in this case via `Cell`.


mwcampbell:
I'll go ahead and merge this, because it's a fairly low-impact change and I've allowed time for it to be reviewed.

PR #63, opened 2021-11-17
Allow lazy initialization of trees

mwcampbell:
When integrating AccessKit with Druid, we'll need to be able to lazily initialize the accessibility tree when the first `WM_GETOBJECT` message on Windows, or equivalent request on other platforms, is received. I think this requirement will be common enough that it's worth supporting directly in AccessKit. So that's what this branch does.

I've implemented this in a backward-compatible way, so it doesn't break the tests and example, but I suppose I really should update them to lazily initialize and update the tree.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,6 +35,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "accesskit_provider"
+version = "0.1.0"
+dependencies = [
+ "accesskit_schema",
+]
+
+[[package]]
 name = "accesskit_schema"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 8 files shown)


mwcampbell:
I believe this is ready for review.

My main concern with this change was that it might be necessary to proactively raise UIA focus change events, particularly when the window gained focus, even if we hadn't yet received a `WM_GETOBJECT` message. But, by studying the relevant code in the open-source WPF repository, testing the modified example with a variety of Windows screen readers, and updating the test infrastructure, I've now convinced myself that it's OK to skip raising focus change events until we've received our first `WM_GETOBJECT` and lazily initialized the tree.

DataTriny:
Very smart! Now the example app is becoming quite complex and I would certainly not understand everything if I had not followed from the start.

It's OK for me, you have some conflicts to fix though.

Issue #37, opened 2021-09-28
UIA provider isn't reliably detected on startup of example program


mwcampbell:
Steps to reproduce:

1. Check out and build [this commit of AccessKit](https://github.com/AccessKit/accesskit/tree/bbf9ccdaa560302e19986e6ca3011e0c4c8a1ed7).
2. On x64 Windows 10 build 19043.1237 (21H1), run NVDA 2021.2.
3. Run the `hello_world` example program under `platforms/windows/examples`.

Expected result: NVDA should say "Hello world window". This indicates that the UIA provider is properly exposed and NVDA is recognizing the window as a UIA window.

Observed: NVDA says "Hello world". This is what it says for a window for which it can't detect a UIA provider. However, if you reverse the order of the steps (start the program and then start NVDA), it says "Hello world window", indicating that it detected the UIA provider.

My analysis: I know that NVDA is behaving as it is because when it calls `UiaHasServerSideProvider` on the newly created window, that function is returning false. But I don't know why `UiaHasServerSideProvider` is returning false for that window. The window procedure handles `WM_GETOBJECT` by creating the UIA provider and passing that along with the `WPARAM` and `LPARAM` to `UiaReturnRawElementProvider`. It doesn't do anything else with the `WPARAM` and `LPARAM`, since I want to let UIA Core handle both UIA and MSAA requests. So I guess there's some kind of race condition where, immediately after the window is created, UIA Core running in NVDA can't connect to UIA Core running in the example program. FWIW, NVDA is 32-bit x86 while the example program is x64.

I have a workaround. If I call `CoInitializeEx` with `COINIT_MULTITHREADED` at the start of the program, then this unexpected behavior doesn't happen. But if I either omit a call to `CoInitializeEx` or call it with `COINIT_APARTMENTTHREADED`, then I get this problem. So for now, I can just use `COINIT_MULTITHREADED`, but since AccessKit is a library aiming for wide adoption, I want it to work in any of these cases if possible. Also, it doesn't seem to matter whether I use `ProviderOptions_UseComThreading` or not, though I'd prefer not to.

I've reported this to one of my former colleagues at Microsoft and his manager, since this looks like it might be a bug in UIA Core.

DataTriny:
I can't reproduce the issue you describe (same Windows build, same Rust architecture, same NVDA version). Running the example from the commit linked above, I get the exact same result as with the next commit which introduce the workaround.
[...]

PR #65, opened 2021-11-18
Handle WM_GETOBJECT while the window is being destroyed

mwcampbell:
This fixes the crash that @DataTriny reported on #37 . It has to be done in the window procedure, since that is what actually manages the lifecycle of the AccessKit manager. So we implement it in our tests and in the example, which provides a model for integrators.


mwcampbell:
Going to go ahead and merge this one, because it's a prerequisite for my next PR, and I don't think anyone else currently involved with the project would be able to meaningfully review this PR.

PR #70, opened 2021-11-24
Use a boxed closure for lazy initialization

mwcampbell:
I was too quick to open my previous PR, trying to make a boxed implementation of `InitTree` work. It didn't actually work for `Box<dyn InitTree>` as I had hoped, because the size of `dyn InitTree` isn't known at compile time.

I don't know why, but `Box<dyn FnOnce>` just works, even though `Box<dyn InitTree>` did not. So I've decided to just use a boxed closure. This bit of dynamism won't actually hurt performance.


DataTriny (approved):
I guess this was ready for review?

It's nice if we can get rid of this very small provider crate, and the initialization code looks cleaner this way.

mwcampbell:
Well, I thought this was for review yesterday, and then I encountered a problem while trying to integrate AccessKit into winit last night. I implemented a solution, but I'm not entirely happy with it. I'm going to try yet another approach on another branch.

mwcampbell:
I now have a solution I'm happy with, and this is ready for review again. The easy default is still a non-thread-safe boxed closure. But `Manager` is now generic, using `Into<TreeUpdate>` instead of a custom trait. This allows us to have an easy default, while still allowing customization for callers that need or want it. A caller can even define its own type that implements `Into<TreeUpdate>` and eliminate the default dynamic dispatch and boxing, as it could before with the old `InitTree` trait.

DataTriny (approved):
Making use of `Into<TreeUpdate>` is very elegant and idiomatic.

PR #72, opened 2021-11-26
Queue events to avoid more deadlocks

mwcampbell:
This is also required for my integration into winit.


platforms/windows/src/manager.rs
```diff
@@ -111,3 +134,49 @@ fn force_init_uia() {
...
+    ///
+    /// The window may receive `WM_GETOBJECT` messages during this call.
+    /// This means that any locks required by the `WM_GETOBJECT` handler
+    /// must not be held when this method is called.
+    ///
+    /// This method should be called on the thread that owns the window.
+    /// It's not clear whether this is a strict requirement of UIA itself,
+    /// but based on the known behavior of UIA, MSAA, and some ATs,
+    /// it's strongly recommended.
+    pub fn raise(self) {
```
DataTriny, line 152: Would it make sense to add `#[must_use]` to this function?

platforms/windows/src/manager.rs
```diff
@@ -111,3 +134,49 @@ fn force_init_uia() {
...
+                    }
+                    .unwrap();
+                }
+            }
+        }
+    }
+}
+
+// We explicitly want to allow the queued events to be sent to the UI thread,
+// so implement Send even though window-rs doesn't implement it for all
```
DataTriny, line 180:
```suggestion
// so implement Send even though windows-rs doesn't implement it for all
```

DataTriny (changes requested):

DataTriny (approved):

PR #76, opened 2021-12-02
Derive more traits on `TreeUpdate`

mwcampbell:
My proof-of-concept integration of AccessKit into [egui](https://github.com/emilk/egui) will be easier if `TreeUpdate` implements `Default`. I realized it will also be helpful to implement `Debug`.


DataTriny (approved):

DataTriny:
I didn't check the whole repo, but I think that we should derive `Debug` for absolutely all of our types if this is not yet the case.

PR #78, opened 2021-12-03
Generate node name from `labelled_by` relationship

mwcampbell:
This is useful for my egui integration, and will likely be useful in other toolkits as well.

### DIFF
--- a/consumer/src/node.rs
+++ b/consumer/src/node.rs
@@ -286,9 +286,17 @@ impl Node<'_> {
 
 #[cfg(test)]
 mod tests {
-    use accesskit::Rect;
+    use accesskit::{Node, NodeId, Rect, Role, StringEncoding, Tree, TreeId, TreeUpdate};
+    use std::num::NonZeroU64;
 
     use crate::tests::*;
+
+    const TREE_ID: &str = "test_tree";
+    const NODE_ID_1: NodeId = NodeId(unsafe { NonZeroU64::new_unchecked(1) });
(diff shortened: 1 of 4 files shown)


consumer/src/node.rs
```diff
@@ -213,11 +213,22 @@ impl<'a> Node<'a> {
...
         } else {
-            None
+            let labelled_by = &self.data().labelled_by;
+            if labelled_by.is_empty() {
+                None
+            } else {
+                Some(
+                    labelled_by
+                        .iter()
+                        .map(|id| self.tree_reader.node_by_id(*id).unwrap().name().unwrap())
```
DataTriny, line 227: Would `filter_map` add a noticeable performance cost here? If not, you could avoid unwrapping. I know that using a nameless element to label another one is stupid, but we already have a ton of `unwrap`.

DataTriny:
I think I would also add a unit test, just to be safe.

mwcampbell:
This is ready to review again.

DataTriny:
I'm OK with the code, but your comment about localization reminded me of a question I had when I first reviewed this: aren't we supposed to expose to ATs the labelled by relationship? AT-SPI allows this and I am pretty sure I saw it for UI Automation as well. Is it a temporary solution, or should we both provide the fully built name plus the labelled by relationships?

mwcampbell:
Yes, we should expose the relationship, to the extent that each platform supports it. Note that on Windows, for instance, UIA's `LabeledBy` property can only hold one element. We should definitely generate the name as well; that's what Chromium does.

DataTriny (approved):

PR #80, opened 2021-12-20
Refactor node transform and bounds using the kurbo crate

mwcampbell:
Now we actually implement transforms, which are now decoupled from the bounding rectangle. Now that we support transforms, the old `offset_container` field of `RelativeBounds` is redundant.

We now use [kurbo](https://github.com/linebender/kurbo) rather than defining our own structures for points, rectangles, and transforms. This reduces the duplicated effort in the Rust graphics ecosystem, and gives us useful functions for manipulating these things.

I'm hoping that an updated version of kurbo with my contributed schemars support will be published to crates.io before I merge this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,6 +7,7 @@ name = "accesskit"
 version = "0.1.0"
 dependencies = [
  "enumset",
+ "kurbo",
  "schemars",
  "serde",
 ]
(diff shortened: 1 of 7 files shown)


mwcampbell:
I'm doing this work now because I'm preparing to implement hit-testing support in AccessKit. My plan is to have AccessKit do its own hit testing rather than running a user-defined callback, to avoid the complications of integrating a synchronous, possibly cross-thread callback into various GUI toolkits.

mwcampbell:
Contrary to what I wrote in a recent commit, I've decided that, at least for now, we don't need a `bounds` method that can return the transformed bounds as a non-rectangular shape. I still think it was a good idea to rename the old `bounds` method to `bounding_box`, though, to indicate that the result of that method will always be a rectangle, regardless of transformations. And we may decide to add the other type of `bounds` method later.

So this PR is now ready for review.

DataTriny (approved):
Let's just hope that kurbo is the right choice.

mwcampbell:
Is there a reason you think kurbo might not be the right choice? It seems pretty lightweight to me.

DataTriny:
I also think that kurbo is a great tool. I was just saying that amongst the GUI crates, only druid is using it right now. Ultimately we will see if other toolkits adopt it as well, or accept to pay the cost of converting to and from their own types.

PR #81, opened 2021-12-20
Handle action requests, starting with focus

mwcampbell:
This defines a trait that providers implement to handle actions requested by assistive technologies, uses that trait to implement UIA's `SetFocus` method, and implements the trait in the Windows example.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -1432,7 +1432,22 @@ pub enum ActionData {
 #[cfg_attr(feature = "serde", serde(deny_unknown_fields))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 pub struct ActionRequest {
-    action: Action,
-    target: NodeId,
-    data: Option<ActionData>,
+    pub action: Action,
+    pub target: NodeId,
+    pub data: Option<ActionData>,
+}
+
+/// Handles requests from assistive technologies or other clients.
+pub trait ActionHandler: Send + Sync {
+    /// Perform the requested action. If the requested action is not supported,
+    /// this method must do nothing.
+    ///
+    /// This method may be called on any thread. In particular, on platforms
+    /// with a designated UI thread, this method may or may not be called
+    /// on that thread. Implementations must correctly handle both cases.
+    ///
+    /// This method may queue the request and handle it asynchronously.
+    /// This behavior is preferred over blocking, e.g. when dispatching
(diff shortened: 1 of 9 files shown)


common/src/lib.rs
```diff
@@ -1433,7 +1433,22 @@ pub enum ActionData {
...
+    /// this method must do nothing.
+    ///
+    /// This method may be called on any thread. In particular, on platforms
+    /// with a designated UI thread, this method may or may not be called
+    /// on that thread. Implementations must correctly handle both cases.
+    ///
+    /// This method may queue the request and handle it asynchronously.
+    /// This behavior is preferred over blocking, e.g. when dispatching
+    /// the request to another thread.
+    fn action(&self, request: ActionRequest);
```
DataTriny, line 1453: I don't think `action` is a nice name for a method. How about `perform_action` or `do_action`?

Also, I know that AT-SPI wants to be notified of the action's success by having its `DoAction` return a `bool`: of course in the context of an asynchronous call this wouldn't make sense, I just wanted to bring this to your attention. Nobody seems to use the result though.

consumer/src/lib.rs
```diff
@@ -41,6 +42,12 @@ mod tests {
     pub const EMPTY_CONTAINER_3_3_IGNORED_ID: NodeId =
         NodeId(unsafe { NonZeroU64::new_unchecked(13) });
 
+    pub struct NullActionHandler();
```
DataTriny, line 45: I didn't know that you could declare an empty struct as an empty tuple struct. I'm not a big fan of it though as its initialization syntax makes it look like a method call. You can simply declare it like so:

```rust
pub struct NullActionHandler;
```

DataTriny (changes requested):
Just a comment about naming, a note about AT-SPI and a minor syntax suggestion.

mwcampbell:
@DataTriny Thanks for the note about AT-SPI. But IMO, encouraging asynchronous handling of actions with no confirmation is important for making it easy to integrate AccessKit into a variety of GUI architectures. There's also an argument from conceptual integrity: arguably the best way to get confirmation that an action was performed is to get a tree update, or other type of event that we might add in the future. After all, that's how a sighted user gets confirmation that a mouse click worked. So I think AT-SPI is misguided on this point, and your implementation should just return true.

I implemented your other suggestions.

DataTriny (approved):

PR #87, opened 2021-12-23
Fill .gitignore; stop tracking Cargo.lock

jacquetc:
Hello,

A little contribution, while I'm trying to see what I can do to help. Good idea for a crate !


mwcampbell:
@jacquetc Thank you for your interest in AccessKit. However, I am tracking Cargo.lock on purpose. The AccessKit project will eventually include bindings for other languages. When it does, we'll offer prebuilt binaries, and tracking Cargo.lock will make sense. Tracking Cargo.lock also lets us avoid surprises in CI. So please restore Cargo.lock and remove that entry from your .gitignore.

jacquetc:
Done :-) I'll not be the first to point that out. It may be useful to add this bit of information somewhere in the doc ?

PR #89, opened 2021-12-24
Support value and range value patterns

mwcampbell:
The UIA value pattern allows access to the value of a text edit control. This will let the screen reader read the current content; full support for text editing will come later. The range value pattern is for progress bars and sliders.

As part of this work, I had to refactor the UIA implementation to use a separate struct for each pattern, to avoid method name conflicts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,6 +54,7 @@ dependencies = [
  "crossbeam-utils",
  "lazy-init",
  "lazy_static",
+ "libc",
  "parking_lot",
  "paste",
  "scopeguard",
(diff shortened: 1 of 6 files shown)


DataTriny (approved):

mwcampbell:
Ready for review again. When I started again this morning, I decided that the concept of numeric value should be explicitly decoupled from range in our own API, even though they're coupled in UIA.

DataTriny (approved):
I also prefer having numeric and range values separated. I shouldn't have approved this in the first place with one job failing anyway...

PR #96, opened 2021-12-29
refactor!: Drop `TreeUpdate::clear`

mwcampbell:
I'm not sure why Chromium has this field, but I don't believe we need it.


mwcampbell:
I'll go ahead and merge this one without review, for two reasons: it's a straightforward cleanup that I don't think anyone will debate, and I want to use this to test the new release process.

PR #102, opened 2022-05-21
Migrate to windows-rs 0.37

str4d:


platforms/windows/src/node.rs
```diff
@@ -703,12 +718,6 @@ macro_rules! patterns {
                 })*
             }
         }
-        $(#[allow(non_snake_case)]
-        impl paste! { [< $base_pattern_id Provider>] } {
-            $(fn $base_property_id(&self) -> Result<$com_type> {
-                self.0.resolve(|resolved| Ok(resolved.$getter().into()))
-            })*
-        })*
```
str4d, line 711: I couldn't think of a nice way to retain the "define the simple things inside the `patterns!` macro" property given `windows` moving to traits, so for the purpose of getting it working, I inlined the simple cases into their respective traits. Technically this means that `$com_type` is now unused inside the macro.

platforms/windows/src/util.rs
```diff
@@ -116,7 +116,7 @@ pub(crate) fn safe_array_from_i32_slice(slice: &[i32]) -> *mut SAFEARRAY {
     safe_array_from_slice(VT_I4, slice)
 }
 
-pub(crate) enum Event {
+pub(crate) enum QueuedEvent {
```
str4d, line 119: I couldn't tell at a glance whether the new `windows::core::Event` type might be able to replace this type, so went with a rename for now.

mwcampbell: The rename is definitely a good solution. Thanks.

platforms/windows/src/node.rs
mwcampbell: Please open an issue for this, then mention the issue in a comment somewhere appropriate in that file. Then we'll see if anyone can figure out an elegant solution.

str4d: Opened #104.

mwcampbell:
@str4d Please update the test harness as you did for the example. I'm aware that our CI doesn't actually run the Windows tests at the moment, but I tried to do so, and they didn't build.

str4d:
Ahh, I missed this because `cargo test` passes, and I thought that ran everything, but if I run `cargo test -p accesskit_windows` I now see the failures. Fixing 😄

str4d:
Okay, this should now be working, and also resolves #104.

mwcampbell:
@str4d Just wondering if you've had a chance to look at the two pending code review comments I left shortly after you opened the PR.

str4d:
I don't see any unresolved issues on this PR. Are they in a review that hasn't been published?

str4d:
The only action items I'm aware of were about the `patterns!` macro and the update to the test harness, both of which I had resolved as of my last comment on May 21st.

DataTriny:
Hello,

Version 0.38.0 was recently published. Is it worth directly upgrading to it? By reading the changelog, it seems to me that this would just be a matter of updating `Cargo.toml`.

Thanks again @str4d for your contribution.

mwcampbell:
@str4d I'm really sorry; I wrote my review comments on May 21, but since this is the first time I've used the GitHub code review feature as a reviewer, I didn't realize I had to submit the review after adding the comments.

mwcampbell:
I've decided that the two review comments I posted aren't blockers. I'll go ahead and merge this, then address those issues myself when I have a chance.

PR #105, opened 2022-05-22
Various documentation fixes and improvements

str4d:


common/src/lib.rs
```diff
@@ -759,9 +779,10 @@ pub struct Node {
     pub busy: bool,
 
     /// The object functions as a text field which exposes its descendants.
+    ///
     /// Use cases include the root of a content-editable region, an ARIA
     /// textbox which isn't currently editable and which has interactive
-    /// descendants, and a <body> element that has "design-mode" set to "on".
+    /// descendants, and a `<body>` element that has "design-mode" set to "on".
```
str4d, line 785: This is what motivated me to make this PR, as the unescaped `<body>` tab [breaks `docs.rs` rendering](https://docs.rs/accesskit/0.3.0/accesskit/struct.Node.html) (scroll down a bit).

common/src/lib.rs
```diff
@@ -378,7 +381,10 @@ pub enum DescriptionFrom {
 
 /// Function that can be performed when a dragged object is released
 /// on a drop target.
-/// Note: aria-dropeffect is deprecated in WAI-ARIA 1.1.
+///
+/// Note: [`aria-dropeffect`] is deprecated in WAI-ARIA 1.1.
+///
+/// [`aria-dropeffect`]: https://www.w3.org/TR/wai-aria-1.1/#aria-dropeffect
```
str4d, line 387: Depending on how you want to present this to downstream users of the `accesskit` crate (who I presume correspond to "authors" in ARIA, vs the "user agents" which seems to match more closely to the platform adapters), we could use:
```suggestion
#[deprecated = "`aria-dropeffect` is deprecated in WAI-ARIA 1.1"]
```

The one downside of this is that the deprecation warnings trigger in this crate as well (this issue is tracked in https://github.com/rust-lang/rust/issues/47219). Most could be worked around with a local `#[allow(deprecated)]`, but I could not find any way to disable these for the `DropEffect` enum specifically other than a global `#![allow(deprecated)]`.

PR #112, opened 2022-07-19
feat(common): Conversion from `NonZeroU64` to `NodeId`

mwcampbell:
We're not sure what size `NodeId` should ultimately be. We may even end up making it configurable through a cargo feature. But we always want to make it easy to convert a `NonZeroU64` to a `NodeId`.

Suggested by @raphlinus.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -19,7 +19,10 @@ use schemars_lib::JsonSchema;
 use serde_lib as serde;
 #[cfg(feature = "serde")]
 use serde_lib::{Deserialize, Serialize};
-use std::{num::{NonZeroU64, NonZeroU128}, ops::Range};
+use std::{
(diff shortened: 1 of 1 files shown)


common/src/lib.rs
```diff
@@ -580,7 +583,7 @@ pub enum StringEncoding {
 
 // This is NonZeroU128 because we regularly store Option<NodeId>.
```
raphlinus, line 584: Suggest replacing this non-doc comment with the following docstring:

```rust
/// The inner type of a [`NodeId`]
///
/// This is currently 128 bits to accommodate UUIDs, but may change or be configurable,
/// to reduce memory usage. It is NonZero because it is commonly used in an `Option`.
```

raphlinus (approved):
Thanks!

PR #114, opened 2022-07-19
refactor!: String indices are always in UTF-8 code units

mwcampbell:
I decided that, given the current state of AccessKit and where most of the interest in this project is coming from, I should optimize for simplicity of Rust code. When AccessKit is eventually used in languages that use UTF-16, the conversion to UTF-8 code units can be done in those languages.

This PR also removes the `testdata` directory, since I haven't been keeping that in sync with the schema anyway.


DataTriny (approved):
Straightforward!

PR #119, opened 2022-07-21
refactor(platforms/windows)!: Simplify the adapter API by always boxi…

mwcampbell:
I decided that the old API was over-complicated, with no real performance benefit. I decided it was time to simplify it before implementing the winit adapter.


mwcampbell:
This change is straightforward enough that I'll just merge it without review.

Issue #117, opened 2022-07-20
winit integration


mwcampbell:
Some popular Rust GUI toolkits and game engines use [winit](https://github.com/Rust-Windowing/winit), so we need an easy way for winit-based applications to use AccessKit.

On Windows, the issue is that winit doesn't currently provide an official way for users to handle custom window messages such as `WM_GETOBJECT`. Possible fixes for this have been discussed on Rust-Windowing/winit#2120, but nothing has yet been implemented. That's not a complaint; I'm sure everyone working on winit is busy with other things. But I need to implement a solution soon. So I'm going to use a Win32 API technique called subclassing to override the window procedure with our own message handler. I'll implement this in the `accesskit_windows` crate, then I can add an `accesskit_winit` crate.

mwcampbell:
I implemented Win32 subclassing in #118.

mwcampbell:
My new winit adapter is in [the winit branch under platforms/winit](https://github.com/AccessKit/accesskit/tree/winit/platforms/winit). I'll submit a PR once #120 is reviewed.

maroider:
The use of a `T: From<ActionRequestEvent>` on the `ActionHandler` impl for `WinitActionHandler<T>` honestly surprised me (in a good way, of course). The API quite reasonable to me, given the current state of Winit's.

mwcampbell:
I just published the accesskit_winit crate on crates.io. So I'll close this, and open separate issues for any problems that arise in real-world usage.

PR #121, opened 2022-07-22
feat(platforms/winit): New winit adapter

mwcampbell:
The example under `accesskit_winit` replaces the winit example in `accesskit_windows`, and now has all of the same behavior as the original example in `accesskit_windows`.


mwcampbell:
I will count @maroider's informal approval on #117 as a review for this PR. I'm eager to get this published.

PR #122, opened 2022-09-04
Enable the serde feature when the schemars feature is turned on

DataTriny:
When enabling the `schemars` feature on the `accesskit` crate, it is necessary to also enable the `serde` feature. Failing to do so will result in an invalid JSON schema being generated due to the fact that serde attributes that are applied on our types won't be present.

Here is a piece of the JSON schema that gets generated when the `serde(rename_all = "camelCase")` attribute is not applied to the `Role` enum:

```json
...
  "definitions": {
    "Role": {
      "description": "...",
      "type": "string",
      "enum": [
        "Unknown",
        "InlineTextBox",
...
```

This PR enables the `serde` feature if the `schemars` feature is enabled, so that it is impossible to accidentally forget to enable both of them.

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -20,5 +20,5 @@ schemars_lib = { package = "schemars", version = "0.8.7", features = ["enumset"]
 serde_lib = { package = "serde", version = "1.0", features = ["derive"], optional = true }
 
 [features]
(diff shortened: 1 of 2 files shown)


DataTriny:
Woops @mwcampbell  clippy is unhappy!

PR #126, opened 2022-09-21
Bump windows-rs dependency

DataTriny:
- Updates `windows-rs` to version `0.40.0` on the `accesskit_windows` and `accesskit_winit` crates.
- Also updates `winit` to version `0.27.3` on the `accesskit_windows` crate (I forgot it previously...).

Hopefully my commit messages are conventional enough that you won't have to rename them! ;)


mwcampbell:
@DataTriny I just noticed one thing: In `platforms/windows/src/tests/mod.rs`, you didn't update the code for registering the window class as you did for the example. The tests still pass, but we should probably be consistent here and get rid of the old way of setting the window class name.

DataTriny:
@mwcampbell you're right. Fixed.

PR #127, opened 2022-09-22
fix(platforms/windows): Use `std::thread::scope` inside tests

DataTriny:


DataTriny:
This obviously bumps MSRV to `1.63` for `accesskit_windows`, but I don't think that is something to worry about right now.

mwcampbell:
Does this affect the MSRV for crates that depend on `accesskit_windows`? If so, then that's a problem. I already had to convince the egui maintainer to increase his MSRV, which he actually enforces, to 1.61, to be compatible with the update to windows-rs 0.37. I'd rather not bump the MSRV again so soon, if this actually affects dependent crates.

DataTriny:
Well, this only affects testing, so dependent crates don't have to compile the updated module. That's also why I labelled this PR as a fix, because I don't think this even deserve a minor version increase.

DataTriny:
@mwcampbell Just to be sure I used [cargo-msrv](https://github.com/foresterre/cargo-msrv) on `accesskit_windows` and it reports `1.61.0`, so we're fine.

PR #130, opened 2022-10-02
fix(common): Write a README specifically for the accesskit crate

### LINKED ISSUE #129: README doesn't show correctly on https://crates.io/crates/accesskit
Right now it just shows:
```
../README.md
```

mwcampbell:
In the last two published versions of the crate, the README wasn't properly sent to the registry because I had set up the crate to use the top-level README via a symbolic link, but then published the crate on Windows, where (at least in my configuration) the symlink doesn't work. I could just publish this crate on Linux, as I did in previous versions, and even automate that step through CI. Ideally we should still do the latter. But I decided that this crate should have its own README, so I wrote one.

fixes #129

### DIFF
--- a/common/README.md
+++ b/common/README.md
@@ -1,1 +1,10 @@
-../README.md
+# AccessKit
+
+This is the shared cross-platform crate for [AccessKit](https://accesskit.dev/). It defines the data structures that represent an accessibility tree, and the trait for handling action requests from assistive technologies.
+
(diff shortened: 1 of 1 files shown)


DataTriny:
So that's why! Yes it's better to have a dedicated README, and the top-level one is a bit out-of-date by now.

DataTriny (approved):

PR #135, opened 2022-10-07
refactor!: Wrap `TreeUpdate` nodes in `Arc`

mwcampbell:
This is another change that's inconvenient for test code, but it eliminates a lot of compiled code for copying around the huge `Node` struct.

### DIFF
--- a/consumer/src/tree.rs
+++ b/consumer/src/tree.rs
@@ -186,7 +186,7 @@ impl State {
 
         fn traverse(state: &State, nodes: &mut Vec<(NodeId, Arc<NodeData>)>, id: NodeId) {
             let node = state.nodes.get(&id).unwrap();
-            nodes.push((id, node.data.clone()));
+            nodes.push((id, Arc::clone(&node.data)));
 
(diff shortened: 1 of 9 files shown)


consumer/src/tree.rs
```diff
@@ -184,9 +184,9 @@ impl State {
     fn serialize(&self) -> TreeUpdate {
         let mut nodes = Vec::new();
 
-        fn traverse(state: &State, nodes: &mut Vec<(NodeId, NodeData)>, id: NodeId) {
+        fn traverse(state: &State, nodes: &mut Vec<(NodeId, Arc<NodeData>)>, id: NodeId) {
             let node = state.nodes.get(&id).unwrap();
-            nodes.push((id, (*node.data).clone()));
+            nodes.push((id, node.data.clone()));
```
DataTriny, line 189: I think I would have used `Arc::clone(&node.data)` here to make it clear that it's not a full clone, but that's only a personal preference I guess.

DataTriny (approved):

PR #138, opened 2022-10-09
fix(consumer): Don't compare node contents during tree update

mwcampbell:
I'm making this change for two reasons:

* Comparing instances of the current, huge `Node` struct is expensive. Eliminating this comparison reduces the size of the Windows example by several kilobytes, whether optimizing for speed or size.
* When I implement the transition to a `Node` trait, comparing across implementations of that trait would be even more expensive, since it would require dynamically dispatched method calls for every property.

This change does eliminate an optimization for immediate-mode GUIs. But honestly, sending a whole tree for every frame isn't good enough for performance anyway; if nothing else, it wastes memory allocations. So I think toolkits will just have to do their own optimizations to avoid sending full trees unnecessarily.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -1198,6 +1198,12 @@ impl Tree {
 /// The sender and receiver must be in sync; the update is only meant
 /// to bring the tree from a specific previous state into its next state.
 /// Trying to apply it to the wrong tree should immediately panic.
+///
+/// Note that for performance, an update should only include nodes that are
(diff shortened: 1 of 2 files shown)


DataTriny:
Indeed, this was one of my concerns. If you plan on releasing a new version before switching to a trait, I would update `TreeUpdate`'s documentation to mention that pushing the same tree over and over is not free. Otherwise it's OK for me.

PR #139, opened 2022-10-09
fix(common): Restore compatibility with Rust 1.61

mwcampbell:
I accidentally increased the MSRV to 1.62 when I used the new feature of deriving `Default` on an enum. That would block integration with egui, so I have to manually implement `Default` instead.


DataTriny (approved):

DataTriny:
Gosh, I forgot this was only introduced three months ago!

PR #145, opened 2022-10-28
feat!: Text range support

mwcampbell:
This allows screen readers and other ATs to provide the expected feedback when editing text on Windows.

The easiest way to try it out with a working application is to try one of the examples in the corresponding [egui branch](https://github.com/mwcampbell/egui/tree/accesskit-text).

The primary missing feature is support for rich text, including text formatting information.

This ought to have tests, but I figured I'd get the implementation into code review.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -970,6 +970,11 @@ pub struct Node {
     /// be non-zero if selecting the line break by itself results in
     /// a visible highlight (as in Microsoft Word), or zero if not
     /// (as in Windows Notepad).
+    ///
+    /// This field is optional. Without it, AccessKit can't support some
+    /// use cases, such as screen magnifiers that track the caret position
+    /// or screen readers that display a highlight cursor. However,
+    /// most text functionality still works without this information.
     #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
     pub character_pixel_lengths: Option<Box<[f32]>>,
 
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -947,6 +947,16 @@ pub struct Node {
     /// should be counted as a single character for the sake of this slice.
     /// When the caret is at the end of such a line, the focus of the text
     /// selection should be on the line break, not after it.
+    ///
+    /// The last inline text box in a text field or document must end
+    /// with a hard line break as described above. If the last line
+    /// is blank, the text box for that line must consist of only
+    /// a hard line break. If the text field is empty, it must consist
+    /// of a single inline text box containing only a hard line break.
+    /// This requirement applies even to text fields that only ever contain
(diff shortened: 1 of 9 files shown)


DataTriny (changes requested):
First pass: egui examples work flawlessly!
We really need an extensive test suite for this one as I'm sure I missed some edge cases.
Outstanding work!

mwcampbell:
The new test suite is not yet comprehensive, but I think it covers enough essential cases that we can consider it good enough for now.

DataTriny (commented):
I am very happy with these changes. I don't see a newline character at the end of text fields anymore.

You still haven't addressed my concern regarding cross-platform newline parsing. If AccessKit should only support \n, then it must at least be documented somewhere.

There is a weird behavior when walking to the next word as the caret is placed at the end of the current word (so usually at the trailing space). I suppose this is a bug in egui, mentioning it here just in case.

Finally, I am not able to use my Braille display routing buttons to move the cursor. When I click above a character, it is only spoken by NVDA but the caret position remains unchanged. I can't see any error on NVDA's log. For reference, I have attached logs demonstrating what is supposed to happen when using the routing keys on a UIA text field (like in the hello_world example, the value is "Arthur", and I clicked on "r").

[NVDA logs.txt](https://github.com/AccessKit/accesskit/files/9989710/NVDA.logs.txt)

mwcampbell:
@DataTriny The fact that line breaks are expected to be either CRLF or LF is actually mentioned in the documentation for `Node::character_lengths`.

egui's behavior for moving by word is indeed not what Windows users and screen readers expect. I'll see if I can get it fixed in egui.

The reason your cursor routing keys aren't working yet is that I forgot to implement the SetTextSelection action in my egui branch. I'll work on that now.

DataTriny:
@mwcampbell Oh sorry, I missed the note on line breaks on the `character_lengths` documentation.

Then I'll wait to test the `SetTextSelection` implementation, and we should be able to merge this quickly!

mwcampbell:
@DataTriny Please test the cursor routing keys with the latest commit on the accesskit-text branch of my egui repo. I had already dispatched the action correctly on the AccessKit site; I just needed to implement it in egui.

mwcampbell:
@DataTriny Good catch. I really should implement hit-testing within text, via the `ITextProvider::RangeFromPoint` method, to complete this feature.

mwcampbell:
@DataTriny OK, NVDA and other ATs should now correctly respond to mouse movement within the text edit. Implementing this turned out to be complicated, so I wrote what I think are thorough tests for this part.

mwcampbell:
@DataTriny I forgot to mention, you should update to the latest commit on the egui accesskit-text branch, so you get the latest Cargo.lock.

DataTriny (approved):
Outstanding work! Let's ship this!

PR #153, opened 2022-11-16
refactor(consumer)!: Eliminate the dependency on `im` due to licensing

mwcampbell:
The `im` crate is under the MPL, which isn't permissive enough for some users. To eliminate the dependency on this crate, we can no longer have a complete snapshot of the old version of the tree when processing tree changes. So we need to introduce the concept of a detached node, and create detached nodes for the old versions of nodes that have been changed or removed.

In the Windows adapter, this change also clarifies the purpose of `NodeWrapper`; that type (now an enum) is now only used when we might need to get a property of a detached node.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17,7 +17,6 @@ name = "accesskit_consumer"
 version = "0.7.1"
 dependencies = [
  "accesskit",
- "im",
  "parking_lot",
 ]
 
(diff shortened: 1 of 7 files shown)


DataTriny (approved):
Unfortunate but necessary. Not a big fan of having to pass all properties to `DetachedNode`, but I don't see a better way to do it.

PR #154, opened 2022-11-17
fix(platforms/winit): Eliminate some problematic indirect dependencies

mwcampbell:
winit's wayland-csd-adwaita feature, which is enabled by default, introduces some dependencies which are rejected by the Bevy game engine's CI. So the accesskit_winit crate now disables this feature. But users of accesskit_winit can still enable this feature if they want.


mwcampbell:
The disabled feature has no affect on winit's public API, and I'm reasonably sure that disabling this feature in the accesskit_winit crate won't impact the integration of the Linux platform adapter. So I'll go ahead and merge this.

PR #157, opened 2022-11-19
Basic macOS setup using `objc2`

madsmtm:
By @mwcampbell, they asked me to do a review and this is easier for me ;)

> The really ugly part of this is the dynamic subclassing (in platforms/macos/src/subclass.rs). But since winit doesn't currently provide a direct way to use a subclass of WinitView, and I have to implement some accessibility methods on the view (starting with accessibilityChildren, but there will be others), I have to do this. And I've got other projects in the works that will require this hack, retrofitting accessibility onto Unity and a GLFW-based GUI. I'm already doing something similar on Windows, though Win32 subclassing is better documented and (I think) more widely used. I'm open to a cleaner solution in winit, but in order to show results immediately, I'm going with this hack for now.


mwcampbell:
Going to close and reopen this under my name, since this is my work, so we have a proper historical record.

PR #158, opened 2022-11-19
feat(platforms/macos): Basic macOS platform adapter

mwcampbell:
The really ugly part of this is the dynamic subclassing (in platforms/macos/src/subclass.rs). But since winit doesn't currently provide a direct way to use a subclass of WinitView, and I have to implement some accessibility methods on the view (starting with accessibilityChildren, but there will be others), I have to do this. And I've got other projects in the works that will require this hack, retrofitting accessibility onto Unity and a GLFW-based GUI. I'm already doing something similar on Windows, though Win32 subclassing is better documented and (I think) more widely used. I'm open to a cleaner solution in winit, but in order to show results immediately, I'm going with this hack for now.


mwcampbell:
Forgot to re-open this as a draft. I've asked @madsmtm to review my usage of `objc2`, but I'm not quite ready to merge the current state of the platform adapter. Things I want to do before I call this PR complete:

* Support lazy tree initialization as I do in the Windows adapter
* Implement hit testing
* Support the increment and decrement actions for sliders and steppers

mwcampbell:
I'll implement text editing support in a separate PR.

platforms/macos/src/subclass.rs
```diff
@@ -0,0 +1,177 @@
...
+        // Cast to a pointer and back to force the lifetime to 'static
+        // SAFETY: We know the class will live as long as the instance,
+        // and we own a reference to the instance.
+        let superclass = unsafe { &*(self.view.class() as *const Class) };
+        self.prev_class = Some(superclass);
+        let mut subclasses = SUBCLASSES.lock();
+        let entry = subclasses.entry(superclass);
+        let subclass = entry.or_insert_with(|| {
+            let name = format!("AccessKitSubclassOf{}", superclass.name());
+            let mut builder = ClassBuilder::new(&name, superclass).unwrap();
```
madsmtm, line 94: Re manual class declaration:

While it'd of course be cleanest if we could use `declare_class!`, I think it's okay if you have to drop down into the manual stuff sometimes. (I'll probably have a bigger document at some point describing this philosophy of "layered safety").

platforms/macos/src/subclass.rs
```diff
@@ -0,0 +1,177 @@
...
+                    hit_test as unsafe extern "C" fn(_, _, _) -> _,
+                );
+            }
+            builder.register()
+        });
+        unsafe { object_setClass(view_ptr as *mut _, (*subclass as *const Class).cast()) };
+    }
+
+    fn uninstall(&self) {
+        let view_ptr = Id::as_ptr(&self.view);
```
madsmtm, line 119: Just FYI: Taking an immutable reference `&self.view` and mutating it like this is only possible because `NSView` effectively contains `ffi::objc_object`, which is `UnsafeCell` internally.

madsmtm (commented):
Overall, looks really good!

I feel I need to include the obligatory warning: `objc2` is still unstable, so there are going to be breaking changes more often in this period. (That said, I will be available for helping you with version upgrades).

Especially the story around main thread safety is quite far from finalized - and that is reflected in your code as well, you have a lot of `unsafe impl Send + Sync` that may not actually be sound.
I suspect a `MainThreadSafe<T>(T)` adapter could be useful to you, might want to experiment with that?

Also, may I recommend at some point adding `#![deny(unsafe_op_in_unsafe_fn)]`? It makes it much easier to narrow down which parts are actually `unsafe` within an `unsafe` function.

Feel free to ask me to clarify anything!

mwcampbell:
@madsmtm Thanks for a thorough review. I'll address the rest of your feedback later today.

DataTriny:
I took a look at the non-MacOS specific changes and they look fine to me.

platforms/macos/src/subclass.rs
```diff
@@ -0,0 +1,177 @@
...
+                    children as unsafe extern "C" fn(_, _) -> _,
+                );
+                builder.add_method(
+                    sel!(accessibilityFocusedUIElement),
+                    focus as unsafe extern "C" fn(_, _) -> _,
+                );
+                builder.add_method(
+                    sel!(accessibilityHitTest:),
+                    hit_test as unsafe extern "C" fn(_, _, _) -> _,
+                );
```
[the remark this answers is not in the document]
madsmtm: Nvmd, you already said in your PR description that that won't be possible

platforms/macos/src/subclass.rs
```diff
@@ -0,0 +1,183 @@
...
+        // and we only use this reference while the instance is alive.
+        let prev_class = unsafe { &*((*view).class() as *const Class) };
+        let mut associated = AssociatedObject::new(adapter, prev_class);
+        unsafe {
+            objc_setAssociatedObject(
+                view as *mut _,
+                associated_object_key(),
+                Id::as_mut_ptr(&mut associated) as *mut _,
+                OBJC_ASSOCIATION_ASSIGN,
+            )
```
madsmtm, line 118: I know that `objc_setAssociatedObject` and `objc_getAssociatedObject` takes and returns an `id`/`objc_object*`, but if you're using `OBJC_ASSOCIATION_ASSIGN`, then it can actually be a pointer to anything - so you didn't need to make `AssociatedObject` a class, you could simply have put it in a `Box`.

platforms/macos/src/subclass.rs
```diff
@@ -0,0 +1,183 @@
...
+unsafe extern "C" fn hit_test(this: &NSView, _cmd: Sel, point: NSPoint) -> *mut NSObject {
+    let associated = associated_object(this);
+    associated.adapter.hit_test(point)
+}
+
+/// Uses dynamic Objective-C subclassing to implement the NSView
+/// accessibility methods when normal subclassing isn't an option.
+pub struct SubclassingAdapter {
+    view: WeakId<NSView>,
+    associated: Id<AssociatedObject, Owned>,
```
madsmtm, line 96: After you've set the associated object, this is no longer the only pointer to the object (and hence `Owned` is wrong)
```suggestion
    associated: Id<AssociatedObject, Shared>,
```

platforms/macos/src/subclass.rs
madsmtm: However, now that you have done so, it can actually be to our benefit:

The code here is a bit dangerous, since if `SubclassingAdapter.associated` (which holds the last retain count to the object) is dropped before the associated object is removed from the view, the view will access a released pointer - of course, this can't happen currently, the implementation of `Drop` ensures that, but still something to keep in mind!

You could consider using `OBJC_ASSOCIATION_RETAIN` instead, which will make the object be retained by the view, which will ensure that it is released only once the view is.

Just a consideration, using `Box` and documenting that the code in `Drop` is necessary for soundness is also fine.

mwcampbell:
@madsmtm I believe I have now addressed all of your feedback. Thanks again.

madsmtm:
> I believe I have now addressed all of your feedback

Yeah, I _think_ it looks good now, but you know, always difficult to say for sure!

> Thanks again

My pleasure, do say so if you need my help with this stuff again

mwcampbell:
If anyone wants to try this adapter with a real GUI toolkit, try [this temporary egui branch](https://github.com/mwcampbell/egui/tree/accesskit-macos). As a reminder, the big missing feature at this point is feedback during text editing.

mwcampbell:
I want to merge this quickly so I can publish it to crates.io and use it in some of my other current work. Since @madsmtm reviewed the safety-critical code in this PR and I addressed his feedback, and @DataTriny reviewed the non-macOS-specific changes, I feel comfortable merging now.

PR #162, opened 2022-11-24
fix(platforms/windows): Re-export more types from windows-rs

mwcampbell:
I forgot that for a clean integration that doesn't require window subclassing, the caller needs more types from windows-rs.


mwcampbell:
Going to go ahead and merge this simple change.

PR #167, opened 2022-11-24
fix(consumer): Allow editable spin buttons

mwcampbell:
On Windows, spin buttons are usually editable, and JAWS assumes that they are. So this PR adds spin buttons to the set of roles that can support text ranges.

### DIFF
--- a/consumer/src/text.rs
+++ b/consumer/src/text.rs
@@ -737,7 +737,10 @@ impl<'a> Node<'a> {
 
     pub fn supports_text_ranges(&self) -> bool {
         let role = self.role();
-        if role != Role::StaticText && role != Role::TextField && role != Role::Document {
+        if !matches!(
+            role,
+            Role::StaticText | Role::TextField | Role::Document | Role::SpinButton
+        ) {
             return false;
         }
         self.inline_text_boxes().next().is_some()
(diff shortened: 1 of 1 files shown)


consumer/src/text.rs
```diff
@@ -737,7 +737,10 @@ impl<'a> Node<'a> {
...
     pub fn supports_text_ranges(&self) -> bool {
         let role = self.role();
-        if role != Role::StaticText && role != Role::TextField && role != Role::Document {
+        if !matches!(
+            role,
+            Role::StaticText | Role::TextField | Role::Document | Role::SpinButton
+        ) {
             return false;
         }
         self.inline_text_boxes().next().is_some()
```
DataTriny, line 746:
```suggestion
        matches!(
            role,
            Role::StaticText | Role::TextField | Role::Document | Role::SpinButton
        ) || self.inline_text_boxes().next().is_some()
```
I think this would be functionally equivalent, but less confusing.

DataTriny (approved):
So there was a bug and none of us catched it before the refactoring!

mwcampbell:
No, I don't believe there was a bug before. But I know it's correct now, so I'll go ahead and merge.

PR #169, opened 2022-11-24
fix: Gracefully handle nodes that only support text ranges some of the time

mwcampbell:
The egui `DragValue` widget will be like this.

BTW, I'm no longer including the parenthesized part in the prefix at the beginning of commit messages, because it doesn't have the purpose that I thought it did in release-please; it just clutters the change logs.


mwcampbell:
Going to go ahead and merge this one. It's a small, simple change, and I tested it together with my work on egui's `DragValue` widget.

PR #174, opened 2022-11-25
fix: Handle views with flipped coordinates

mwcampbell:
I didn't realize that on macOS, a view's coordinates can be flipped so that y is top-down (as it is on other platforms). druid-shell and its successor glazier do this, so we need to handle it.


DataTriny (approved):
Code-wise it looks OK.

PR #179, opened 2022-11-28
refactor!: Move lazy initialization from the core platform adapter to the caller

mwcampbell:
While working on the new integration of AccessKit into [glazier](https://github.com/linebender/glazier), I'm trying to make my API additions feel integral to the design of glazier, not bolted on. In particular, this means that the logical place for applications to return an initial accessibility tree is in a method on glazier's `WinHandler` trait. But this trait has several methods that take a mutable reference to `self`, which means it's not practical for the trait object to be reference-counted. That means there isn't a clean way to get a reference to that trait object in the tree source/factory closure that we currently pass to `Adapter`. I expect that future integrations would have similar challenges with that existing design in AccessKit. We already know that the Windows hello_world example had to be somewhat contorted to deal with it.

The problem is that the closure that we currently pass to `Adapter` has arbitrary lifetime, may be called at any time, and may only refer to variables that it has closed over from its parent function. But in a clean AccessKit integration where the caller can implement its own handling of the `WM_GETOBJECT` message on Windows or the equivalent on other platforms, lazy adapter initialization only has to be done in specific contexts, and in those contexts, the caller already has access to specific state such as the glazier `WinHandler` trait object.

So I've decided that the core platform adapters should not have built-in support for laziness; instead, it should be the caller's responsibility to lazily initialize the adapter. The Windows hello_world example now shows the best way to do this in a clean AccessKit integration.

Note that the hello_world example and the Windows test harness take different approaches to laziness. The approach in the hello_world example is meant to be a good example of how clean integrations should do it, while the approach taken in the test harness is messier but convenient given the need to pass more dynamic data to the test window.

There's also an unfortunate wart in the new design on Windows: the `UiaInitMarker` struct. The trouble is that we now expect that `Adapter` will be created inside the `WM_GETOBJECT` handler, but that is exactly when we don't want to initialize UIA. I can't think of a better way. At least with `UiaInitMarker`, we try to guide the caller into handling UIA initialization the right way.

I mentioned clean integrations above. We also have subclassing, which is clearly dirty. A subclassing adapter has to be initialized early, before the window is shown; this means it has to handle lazy initialization of the underlying adapter. I decided that both `SubclassingAdapter` structs (on Windows and macOS) should take full responsibility for the life cycle of the underlying `Adapter`. While I was at it, I refactored the API on Windows to be a closer match for what I recently implemented on macOS. I also eliminated the need for the closure to be boxed, on both platforms as well as in the winit adapter.

@DataTriny I've thought a little about how we might achieve laziness in the AT-SPI adapter. Clearly the adapter has to be created when the window is created, so that if AT-SPI is enabled, the adapter can create its bus and register itself. And while your `Adapter::new` takes a closure, it always calls the closure immediately, so there's no real laziness. Here's what I think you should do:

1. The closure parameter should be `impl FnOnce() -> TreeUpdate`. Without being a boxed trait object, and without the `'static` lifetime bound, the closure doesn't have to live any longer than the call to the `new` function. I can work with that in a scenario like the one in glazier that prompted this refactor.
2. The `new` function should return `Option<Self>`, and that result should be `None` in the case where `Adapter::atspi_bus` would currently be `None`. So, if there's no AT-SPI bus, then `Adapter::new` returns `None` and the provided closure is never called. This, I think, is as lazy as we can get.


mwcampbell:
To see how this refactor enables the integration in glazier, look at [this branch](https://github.com/mwcampbell/glazier/tree/accesskit). Notice how straightforward the new example of using AccessKit with glazier is.

DataTriny (approved):
Nothing bad caught my eyes.

PR #181, opened 2022-12-01
fix: Expose which accessibility selectors are actually allowed for a particular node

mwcampbell:
I think this is a good thing to add before implementing the `NSAccessibility` text methods, which an AT shouldn't try to call on nodes that don't support text ranges. Since I've chosen to have a single Objective-C class for all types of AccessKit nodes, it's useful to let ATs know which methods are and are not supported for a particular node.

I know this method is being used, because before I added `accessibilityRoleDescription` to the list of selectors that are allowed by default, VoiceOver didn't speak some role descriptions, like "slider", "stepper", and "button". Maybe we'll eventually discover more selectors that need to be on the default allow list.

### DIFF
--- a/platforms/macos/src/node.rs
+++ b/platforms/macos/src/node.rs
@@ -18,7 +18,8 @@ use objc2::{
     foundation::{NSArray, NSCopying, NSNumber, NSObject, NSPoint, NSRect, NSSize, NSString},
     msg_send_id, ns_string,
     rc::{Id, Owned, Shared},
-    ClassType,
+    runtime::Sel,
+    sel, ClassType,
 };
 use std::{
     ptr::null_mut,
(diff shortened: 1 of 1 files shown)


spencerudnick:
Have you checked if this works in practice?

In my experience, `isAccessibilitySelectorAllowed:@selector(accessibilityRoleDescription)` does not actually work as desired.

The only way I've found to properly support custom role descriptions is to _not_ implement `accessibilityRoleDescription` and instead use `setAccessibilityRoleDescription` on the `NSAccessibilityElement*`.

spencerudnick:
In general, I've found VoiceOver's use of `isAccessibilitySelectorAllowed` to be inconsistent, and not reliable. It's why the direction I took was to define separate sub-interfaces of `NSAccessibilityElement<NSAccessibilityElement>`.

mwcampbell:
I found that overriding `accessibilityRoleDescription` is working fine so far. I'll be submitting support for custom role descriptions as a separate PR. The trick, I think, is to call the method on the superclass if we don't have our own role description.

Still, I guess I should wait and merge this PR only once I've confirmed that this approach works for the text-related selectors.

mwcampbell:
So far, my current approach, having a single node class and implementing `isAccessibilitySelectorAllowed:`, is working well, even after introducing text editing, so I'll merge this PR now.

PR #182, opened 2022-12-01
docs: Fix outdated documentation for `TreeUpdate`

mwcampbell:
AccessKit's tree update code is somewhat more lenient than Chromium's seemed to be. I think the new docs are also clearer than what I took from Chromium.


DataTriny (approved):
Yes, much clearer.

PR #184, opened 2022-12-02
feat: Automatically get button and link labels from descendants

mwcampbell:
This makes it easier to implement accessibility in some GUI frameworks that build up widgets from simpler components.


DataTriny (approved):
I find the iterator quite elegant. But now that we have that, is the plan to expose this through the adapters?
I was never a fan of how we build node names.

mwcampbell:
I believe we have no choice but to build the node name as we do on Windows. We'll probably end up exposing the labelled-by relation on Mac though.

PR #185, opened 2022-12-02
fix: Correctly apply the DPI scale factor to coordinates

mwcampbell:
I incorrectly assumed that the relevant accessibility methods used physical (DPI-dependent) coordinates. A Retina Mac user noticed that the VoiceOver highlight was wrong. I already asked him to test this fix.


mwcampbell:
Going to go ahead and merge this one without review. I want to publish new AccessKit crates before the initial egui AccessKit PR is merged, and this is an important fix.

PR #186, opened 2022-12-02
fix: Expose static text as the value rather than the title on macOS

mwcampbell:
macOS expects the text content of a static text control to be exposed as the value rather than the title. The immediate effect of this fix is that VoiceOver no longer adds "text" to the end of the text. This will probably become more important once we properly expose the labelled-by relation.

### DIFF
--- a/platforms/macos/src/node.rs
+++ b/platforms/macos/src/node.rs
@@ -271,18 +271,29 @@ impl<'a> NodeWrapper<'a> {
     // and figure out how this is different in the macOS 10.10+ protocol
 
     pub(crate) fn title(&self) -> Option<String> {
+        let state = self.node_state();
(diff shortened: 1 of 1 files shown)


platforms/macos/src/node.rs
```diff
@@ -271,19 +271,30 @@ impl<'a> NodeWrapper<'a> {
...
+        if let Some(value) = state.value() {
             return Some(Value::String(value.into()));
         }
-        if let Some(value) = self.node_state().numeric_value() {
+        if let Some(value) = state.numeric_value() {
             return Some(Value::Number(value));
         }
+        if state.role() == Role::StaticText {
+            if let Some(name) = self.name() {
+                return Some(Value::String(name));
```
spencerudnick, line 295: 👍

spencerudnick (approved):

PR #187, opened 2022-12-02
fix: Don't expose the window title in our root element on macOS

mwcampbell:
If we expose the window title on the group element that we expose for the root of the window, VoiceOver behavior is broken. On macOS, VoiceOver gets the window title from the Cocoa `NSWindow`.


DataTriny (approved):
It's unfortunate... The fix is simple so that's fine.

spencerudnick (approved):

PR #191, opened 2022-12-14
feat: Text support on macOS

mwcampbell:
This should be on par with the Windows adapter, meaning support for single-line and multi-line edit controls, but without exposing text formatting attributes or other rich text features. To test with a real toolkit implementation, use [this egui branch](https://github.com/mwcampbell/egui/tree/accesskit-macos-basic-text). egui itself required only one minor fix, which is already merged; the existing integration in egui that was tested with the Windows adapter was mostly fine already.

### DIFF
--- a/platforms/macos/src/event.rs
+++ b/platforms/macos/src/event.rs
@@ -115,6 +115,15 @@ impl TreeChangeHandler for EventGenerator {
                 notification: unsafe { NSAccessibilityValueChangedNotification },
             });
         }
+        if old_wrapper.supports_text_ranges()
+            && new_wrapper.supports_text_ranges()
+            && old_wrapper.raw_text_selection() != new_wrapper.raw_text_selection()
+        {
+            self.events.push(QueuedEvent::Generic {
+                node_id,
+                notification: unsafe { NSAccessibilitySelectedTextChangedNotification },
+            });
+        }
     }
 
     fn focus_moved(&mut self, _old_node: Option<&DetachedNode>, new_node: Option<&Node>) {
--- a/platforms/macos/src/node.rs
+++ b/platforms/macos/src/node.rs
@@ -421,6 +421,12 @@ declare_class!(
             .unwrap_or_else(null_mut)
         }
 
+        #[sel(setAccessibilityValue:)]
+        fn set_value(&self, _value: &NSObject) {
+            // This isn't yet implemented. See the comment on this selector
+            // in `is_selector_allowed`.
+        }
+
         #[sel(accessibilityMinValue)]
         fn min_value(&self) -> *mut NSNumber {
             self.resolve(|node| {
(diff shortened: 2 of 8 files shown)


mwcampbell:
I'm aware that the new functions for converting to and from indices are O(n), where n is the number of lines or even characters. I know this won't scale to huge chunks of text, but I've decided not to bother with caching or the like until we know that we need it. After all, cache invalidation is one of the hardest problems in computer science.

TheQuinbox:
I don't even think AppKit bothers with caching as far as that goes, because when in a large text field, pressing command+down to try to jump to the end causes the application to lock up for a good while. If not lock up, at the very least, make VoiceOver think it went not responding. I'm unsure if it also blows up for sighted users, but I doubt it.
This freeze period is longer on Intel Macs, but still happens on M1's. If this proves to be a problem you can look into it, but I wouldn't stress it.

DataTriny (commented):
Sounds good to me, but I'd prefer if someone with MacOS development knowledge could have a look.

mwcampbell:
@spencerudnick Can you please review this one?

platforms/macos/src/event.rs
```diff
@@ -115,6 +115,15 @@ impl TreeChangeHandler for EventGenerator {
                 notification: unsafe { NSAccessibilityValueChangedNotification },
             });
         }
+        if old_wrapper.supports_text_ranges()
+            && new_wrapper.supports_text_ranges()
+            && old_wrapper.raw_text_selection() != new_wrapper.raw_text_selection()
```
spencerudnick, line 120: This is a nice way to do it!

platforms/macos/src/node.rs
```diff
@@ -513,6 +538,240 @@ declare_class!(
             true
         }
 
+        #[sel(accessibilityNumberOfCharacters)]
+        fn number_of_characters(&self) -> NSInteger {
+            self.resolve(|node| {
+                if node.supports_text_ranges() {
+                    node.document_range().end().to_global_utf16_index() as _
```
spencerudnick, line 517: I'm unfamiliar with casting in rust, is this `_` similar to typescript's `any`?

platforms/macos/src/node.rs
```diff
@@ -528,6 +787,25 @@ declare_class!(
...
+                    || selector == sel!(accessibilityRangeForPosition:)
+                    || selector == sel!(accessibilityStringForRange:)
+                    || selector == sel!(accessibilityAttributedStringForRange:)
+                    || selector == sel!(accessibilityFrameForRange:)
+                    || selector == sel!(accessibilityLineForIndex:)
+                    || selector == sel!(accessibilityRangeForIndex:)
+                    // This adapter doesn't yet actually support the SetValue
+                    // action, but we must claim to support it in order
+                    // to get the expected VoiceOver behavior in edits.
+                    || selector == sel!(setAccessibilityValue:)
```
spencerudnick: Interesting that this is necessary, thanks for documenting it.

Have you tested what happens if you try to set the value via something like Accessibility Inspector?

platforms/macos/src/node.rs
mwcampbell: No, the `_` uses type inference to cast to whatever type is necessary. In this case, it's shorthand for casting to `NSInteger`.

spencerudnick (approved):
Nice work, everything looks good to me. Though I had one question about claiming to support `setAccessibilityValue:` without actually implementing it.

PR #192, opened 2022-12-15
fix: More reliable handling of the edge case for wrapped lines

mwcampbell:
I discovered that NVDA was unable to move forward to the next line of text when the current line was wrapped (i.e. ended with a soft line break). This revealed some flaws in my current approach to setting the bias (also called affinity) for text positions on the edge of a soft line break. The new approach is more explicit. This fix also ensures that the start of a non-degenerate range is never on the end of a previous line. I added test coverage for all of this, in the tests for the consumer crate. I should still write tests for the Windows adapter sometime in the future.

### DIFF
--- a/consumer/src/text.rs
+++ b/consumer/src/text.rs
@@ -254,11 +254,11 @@ impl<'a> Position<'a> {
     pub fn to_line_index(&self) -> usize {
         let mut pos = *self;
         if !pos.is_line_start() {
-            pos = pos.backward_by_line();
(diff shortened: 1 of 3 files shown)


DataTriny (approved):

mwcampbell:
This PR had some overlap with the macOS text PR, so I merged that one first, then rebased and fixed this one. I believe the fixes I did after the rebase are straightforward enough to not need additional review.

PR #193, opened 2022-12-15
fix: Correct broken UIA method implementation that was incompatible with Windows 11 ATs

### LINKED ISSUE #189: TextField characters not read by screen reader on Windows 11
While integrating AccessKit with my own GUI lib, using the egui integration as reference, I noticed that narrator (the screen reader I am using for testing) is not working correctly with a textbox (TextField role with InlineTextBox child). The characters are not read when shifting the cursor with the arrow keys or when selecting the text. To make sure this wasn't a problem on my end I ran the egui example and found the same problem.

I don't know if this problem is specific to Windows 11 but I can check on a Windows 10 machine tomorrow.

mwcampbell:
This prevented Narrator on Windows 11 from reading characters, selection changes, and possibly other things. It might have also affected other assistive technologies that use the same newer UI Automation API that was introduced in Windows 11 and is now being used by Narrator.

fixes #189

### DIFF
--- a/platforms/windows/src/text.rs
+++ b/platforms/windows/src/text.rs
@@ -570,6 +570,6 @@ impl ITextRangeProvider_Impl for PlatformRange {
 
     fn GetChildren(&self) -> Result<*mut SAFEARRAY> {
         // We don't support embedded objects in text.
-        Ok(std::ptr::null_mut())
+        Ok(safe_array_from_com_slice(&[]))
     }
 }


geom3trik:
Tested on my Windows 11 machine (version 22000) and is working correctly.

DataTriny:
@mwcampbell CI is failing. Can you please fix clippy?

DataTriny (approved):
Fine. Maybe rebase just to be safe.

PR #194, opened 2022-12-16
fix: Provide fallback property implementations for the window root

mwcampbell:
UIA itself normally does this automatically. But for an AT that uses the UIA Remote Operations API introduced in Windows 11, such as the Windows 11 version of Narrator, this fallback isn't available for providers that don't use COM threading, such as ours. As a result, Windows 11 Narrator was unable to find the window context for the current focus if something other than the window is immediately focused. We don't want to switch to COM threading, since the ability to run UIA methods independently of the UI thread is a key benefit of the AccessKit architecture. We also don't want to require integrators to provide the window title as the name of the root node if we can avoid it. So, this PR implements our own fallback. It also modifies the Windows example and test code to clarify that the window title doesn't need to be explicitly provided in the AccessKit tree.

### DIFF
--- a/platforms/windows/examples/hello_world.rs
+++ b/platforms/windows/examples/hello_world.rs
@@ -84,7 +84,6 @@ fn get_initial_state() -> TreeUpdate {
     let root = Arc::new(Node {
         role: Role::Window,
         children: vec![BUTTON_1_ID, BUTTON_2_ID],
-        name: Some(WINDOW_TITLE.into()),
         ..Default::default()
     });
     let button_1 = make_button(BUTTON_1_ID, "Button 1");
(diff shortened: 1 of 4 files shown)


mwcampbell:
Abandoning this PR. We need a more comprehensive solution for the underlying problem, aprticularly on Windows 11 build 22000.

mwcampbell:
Looks like I just needed to do the same for the `NativeWindowHandle` property.

geom3trik:
Tested on my Windows 11 machine (version 22000) and is working correctly.

DataTriny:
This would make our Unity plugin a bit simpler.

> We also don't want to require integrators to provide the window title as the name of the root node if we can avoid it.

This behaviour is very Windows specific right?

mwcampbell:
@DataTriny They also don't have to provide the window title on macOS, because VoiceOver gets it from the parent window, not the content view. I'll let you decide if we can make this fallback behavior work on Linux, but I'd like to try.

mwcampbell:
@DataTriny Regardless of whether the window title fallback is practical to implement on Linux, what do you think of this PR? Once this PR is merged, I can release new AccessKit crates with all the recent fixes plus macOS text editing.

DataTriny (approved):

DataTriny:
@mwcampbell Apart from this bit of uncertainty, I'm OK with these changes.

PR #195, opened 2022-12-17
chore: Fix clippy lints and make CI more comprehensive

mwcampbell:
Rust 1.66 introduced a couple of new lints. There's also one from the Windows example that wasn't caught by the CI.


DataTriny:
I think it's not the first time that the CI misses something from an example. I wonder whether examples are actually fully checked in our CI.

mwcampbell:
OK, added the `--all-targets` option in CI and fixed a few more lints in Windows tests.

DataTriny (approved):

PR #196, opened 2022-12-20
feat: Live regions on macOS

mwcampbell:
Cocoa accessibility doesn't seem to have live regions as such, so we post announcement notifications instead. This is also what Chromium and Firefox do.

### DIFF
--- a/platforms/macos/src/event.rs
+++ b/platforms/macos/src/event.rs
@@ -3,16 +3,30 @@
 // the LICENSE-APACHE file) or the MIT license (found in
 // the LICENSE-MIT file), at your option.
 
-use accesskit::NodeId;
+use accesskit::{Live, NodeId};
 use accesskit_consumer::{DetachedNode, FilterResult, Node, TreeChangeHandler, TreeState};
-use objc2::foundation::NSString;
+use objc2::{
+    foundation::{NSInteger, NSMutableDictionary, NSNumber, NSString},
+    msg_send,
+    rc::Id,
+    Message,
+};
 use std::rc::Rc;
 
 use crate::{
     appkit::*,
     context::Context,
-    node::{filter, NodeWrapper},
+    node::{filter, filter_detached, NodeWrapper},
 };
+
+// Workaround for https://github.com/madsmtm/objc2/issues/306
+fn set_object_for_key<K: Message, V: Message>(
+    dictionary: &mut NSMutableDictionary<K, V>,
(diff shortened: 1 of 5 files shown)


mwcampbell:
@madsmtm Please review my workaround for <https://github.com/madsmtm/objc2/issues/306>, or tell me if I have overlooked an obvious solution.

platforms/macos/src/event.rs
```diff
@@ -3,17 +3,31 @@
...
+    node::{filter, filter_detached, NodeWrapper},
 };
 
+// Workaround for https://github.com/madsmtm/objc2/issues/306
+fn set_object_for_key<K: Message, V: Message>(
+    dictionary: &mut NSMutableDictionary<K, V>,
+    value: &V,
+    key: &K,
+) {
+    let _: () = unsafe { msg_send![dictionary, setObject:value forKey:key] };
```
madsmtm, line 28: Nit: Add comma between arguments

```suggestion
    let _: () = unsafe { msg_send![dictionary, setObject: value, forKey: key] };
```

platforms/macos/src/event.rs
```diff
@@ -44,6 +73,40 @@ impl QueuedEvent {
...
+                let mut user_info = NSMutableDictionary::new();
+                let text = Id::into_super(NSString::from_str(&text));
+                set_object_for_key(&mut user_info, &*text, unsafe {
+                    NSAccessibilityAnnouncementKey
+                });
+                let priority = Id::into_super(Id::into_super(NSNumber::new_isize(priority)));
+                set_object_for_key(&mut user_info, &*priority, unsafe {
+                    NSAccessibilityPriorityKey
+                });
+                let user_info = Id::into_super(user_info);
```
madsmtm, line 100: Nit: I'm fairly certain you can use auto-deref like this to avoid having to call `into_super` so much.

```suggestion
                let mut user_info: Id<NSMutableDictionary<NSString, NSObject>, _> = NSMutableDictionary::new();
                let text = NSString::from_str(&text);
                set_object_for_key(&mut user_info, &*text, &*unsafe {
                    NSAccessibilityAnnouncementKey
                });
                let priority = NSNumber::new_isize(priority);
                set_object_for_key(&mut user_info, &*priority, &*unsafe {
                    NSAccessibilityPriorityKey
                });
```

madsmtm (approved):
The workaround looks fine

PR #203, opened 2023-01-06
fix: Make `Node::filtered_parent` recursive as it was meant to be

mwcampbell:
I originally implemented this correctly, but then I got in a hurry and let clippy automatically fix a warning without double-checking the fix. And since there was no test coverage for this method, the bug got through. So I added a test as well.


mwcampbell:
This is straightforward enough and critical enough that I'll merge it without further review.

PR #207, opened 2023-02-02
Update winit to 0.28

emilk:
I also removed `Cargo.lock` from git, since this is a library (see https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html). Let me know if I went to far here!


emilk:
Sure, no problem!

platforms/winit/Cargo.toml
```diff
@@ -23,6 +27,3 @@ accesskit_macos = { version = "0.4.2", path = "../macos" }
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
 accesskit_unix = { version = "0.1.1", path = "../unix" }
-
-[dev-dependencies]
-winit = "0.27.2"
```
emilk, line 28: It is already a proper dependency

mwcampbell:
The reason I listed winit again under dev-dependencies was so our example could run with all default features, including Wayland client-side decorations where necessary. Disabling that feature in the base dependency on winit is no longer as necessary, now that winit doesn't depend on crossfont by default, but I'd rather keep it that way.

emilk:
Well, serves me right for thinking I know better :)

mwcampbell:
No worries. You _are_ a more experienced Rust dev than I am. Guess I should have added a comment there.

mwcampbell:
OK, accesskit_winit v0.9.0 is published. I realize this is necessary to unblock the update to winit 0.28 in egui.

emilk:
Thanks for the quick response! 🙏

PR #209, opened 2023-02-03
fix: Don't force winit's X11 and Wayland features to be enabled

mwcampbell:
Bevy prefers to keep these features enabled by default and let the user choose which of them to enable. Since the Unix backend in `accesskit_winit` didn't actually need to use anything specific to X11 or Wayland, we can disable these features too.

### DIFF
--- a/platforms/winit/Cargo.toml
+++ b/platforms/winit/Cargo.toml
@@ -13,7 +13,7 @@ edition = "2021"
 [dependencies]
 accesskit = { version = "0.8.1", path = "../../common" }
 parking_lot = "0.12.1"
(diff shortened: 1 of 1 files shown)


mwcampbell:
@DataTriny Do you plan to make any additions to the Unix backend in the winit adapter that will require access to the platform-specific (i.e. X11 and/or Wayland) APIs of winit? If so, then we should discuss how to do that while not forcing those features to be enabled, as requested by the Bevy team. If not, then I think this PR is safe to merge.

DataTriny (approved):
@mwcampbell I don't currently see any addition to `accesskit_winit` that would require these features. I should have removed them from the start.

PR #212, opened 2023-02-09
refactor!: Move thread synchronization into platform adapters; drop parking_lot

mwcampbell:
This PR has the following practical benefits:

* AccessKit no longer depends on parking_lot, except indirectly in the Unix adapter (via async-broadcast-channel which is used by zbus).
* Because the consumer crate no longer does thread synchronization itself, we're a step closer to making it compatible with no-std environments, e.g. embedded devices and game consoles.
* Because thread synchronization is now the responsibility of the platform adapters, we don't need any on macOS, since everything happens on the main thread on that platform. The same will likely be the case on iOS. That means that we don't need to use the somewhat inefficient pthread-based synchronization primitives on Apple platforms, but we don't need to use something with some global overhead, like parking_lot, either.
* Platform-specific implementations of `ActionHandler` only need to implement `Send` and `Sync` if the platform actually requires it, which isn't the case for macOS. This will let us optimize the AccessKit integration in glazier's macOS backend a bit. The same will also be true for iOS and WebAssembly when we get to those platforms.
* Speaking of WebAssembly, accesskit_winit now compiles for that target, though of course it uses the null backend for now.

We're close to being able to merge `Tree` and `TreeState` in the consumer crate, but we can't do that without reworking the API for handling tree changes, so I'm leaving that alone for now.

I needed to rework the Windows and Unix adapters to use a top-level context struct, as the macOS adapter already did. I was planning to do that at some point anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,7 +32,6 @@ name = "accesskit_consumer"
 version = "0.13.0"
 dependencies = [
  "accesskit",
- "parking_lot",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,7 +42,6 @@ dependencies = [
  "accesskit_consumer",
  "objc2",
  "once_cell",
- "parking_lot",
 ]
 
 [[package]]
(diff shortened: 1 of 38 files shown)


mwcampbell:
Also, this shouldn't require AccessKit users to do anything other than bump version numbers.

DataTriny (approved):

PR #214, opened 2023-02-18
feat: Feature-gate the Unix adapter in accesskit_winit

mwcampbell:
The current effort to integrate AccessKit in the Bevy game engine is meeting with some resistance because of the number of new dependencies added by the AccessKit Unix adapter. So, as a compromise, this PR allows just the Unix adapter to be turned off with a feature in accesskit_winit. If the feature is disabled, then accesskit_winit will use the null adapter on Unix platforms, as it does for unsupported platforms.

### DIFF
--- a/platforms/winit/Cargo.toml
+++ b/platforms/winit/Cargo.toml
@@ -10,6 +10,9 @@ repository = "https://github.com/AccessKit/accesskit"
 readme = "README.md"
 edition = "2021"
 
+[features]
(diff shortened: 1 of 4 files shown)


DataTriny (commented):
You should probably update `platforms/winit/examples/simple.rs` (lines 146 to 153) to stay consistent.

DataTriny (approved):

PR #216, opened 2023-02-19
fix: Set `repr(transparent)` on `NodeId`

mwcampbell:
This should help with FFI.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -644,9 +644,9 @@ pub type NodeIdContent = NonZeroU128;
 
 /// The stable identity of a [`Node`], unique within the node's tree.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
-#[repr(transparent)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "schemars", derive(JsonSchema))]
+#[repr(transparent)]
 pub struct NodeId(pub NodeIdContent);
 
 impl From<NonZeroU64> for NodeId {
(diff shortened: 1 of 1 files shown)


DataTriny:
@mwcampbell If you really want to tackle this separately from a following C bindings PR, then check out my [branch](https://github.com/DataTriny/accesskit/tree/c_bindings) because there are a lot of other places that require attention.

common/src/lib.rs
```diff
@@ -644,6 +644,7 @@ pub type NodeIdContent = NonZeroU128;
 
 /// The stable identity of a [`Node`], unique within the node's tree.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+#[repr(transparent)]
```
DataTriny, line 647: Currently, on all types that have a `#repr(...)] attribute, it comes last, just before the actual struct declaration. I'd suggest moving it after the `#cfg|...]` attributes to be consistent.

DataTriny (changes requested):

mwcampbell:
I decided to go with `repr(u8)` on the enums rather than `repr(C)`, because most of our enums have a very small number of values, and even `Role` seems unlikely to exceed 256 values in the future, considering how comprehensive the set of roles from Chromium is.

DataTriny:
@mwcampbell I indeed agree with you. I just forgot that I didn't touch on `ActionData` yet because I wasn't sure on how it would be represented on the C side. But now that I've read a bit more about FFI, I think `#[repr(C)]` makes more sense here.

mwcampbell:
Do you mean we should use `repr(C)` in all the places where I just added `repr(u8)`? If so, why? I don't think an enum with fields, like `ActionData`, can be directly represented in the C API. Not sure what to do about that.

DataTriny:
I agree with all the changes you made. I was just referring to `ActionData`. I think it can be represented with `#[repr(C)]` since all the variants have a single field so it's kind of a tagged union. I can deal with this in my C bindings PR if you are not comfortable introducing this change right now.

DataTriny (approved):
Nice, thanks for that! Don't forget to update the commit title to mention the other changes.

PR #217, opened 2023-02-20
fix: Update atspi dependency

mwcampbell:
@DataTriny Please make sure this doesn't break anything in the Unix adapter. If not, let's get this out quickly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -229,9 +229,9 @@ dependencies = [
 
 [[package]]
 name = "atspi"
-version = "0.10.0"
+version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c5490b0219179887bd331bbbb25d9a940daea89bf59af502505143f864fd301b"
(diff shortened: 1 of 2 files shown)


DataTriny (approved):
The winit example still runs fine! Doesn't seem to be a win in terms of dependency count though...

mwcampbell:
I caught the extra dependencies (a bit late) and reported the issue. Tait turned around a fix quickly. Please check again; should be a win now.

DataTriny:
@mwcampbell I know about the two extra deps, I was planning on submitting a PR today but you were quicker. The winit example used to report 200 dependencies and now it's at 201. I don't think we should take the time to investigate this now so let's land this and deal with the dependency issue later. At least binary size decreased and fore me this is the most important.

DataTriny:
Oh maybe I wasn't clear here. Indeed the dependency count dropped by bumping to latest `atspi`. But the count seems to have roughly stayed the same between the time I tried yesterday with latest `main` (before your PR) and today.

PR #220, opened 2023-02-20
fix: Update windows-rs to 0.44

mwcampbell:
The immediate motivation is a CI failure in the Bevy AccessKit PR. While Bevy is indirectly depending on windows-rs 0.43, I was told we should go ahead and update to 0.44.


DataTriny (approved):

DataTriny:
To me their job will still find two versions of windows-rs unless they update to 0.44 as well. Crossing fingers.

PR #230, opened 2023-03-19
feat: Add C bindings

### LINKED ISSUE #206: Are there C API (code bindings) and pre-built binaries?
I would like to link AccessKit to my Golang app.

DataTriny:
Closes #206 

I think this is now good enough for others to have a look.

I have only compiled the macOS module, but I haven't used it from C, so the headers might now be valid on that platform.

@mwcampbell I'll need you to add a GitHub secret named `RELEASE_PLEASE_TOKEN` to the repo so the `release-please` workflow can trigger the new one. It must contain a personal access token. On my tests I've given full repo and workflow permissions, but the later might not be needed.


mwcampbell:
I suggest that we decide on a C formatting style for both our header file and our C examples and use clang-format to enforce it. I don't particularly care which one we use.

mwcampbell:
In the Windows C example, in `inner_window_state_build_root`, I think it would be better to use `accesskit_node_builder_push_child` for all three children, even though that's not a direct translation of what I did in Rust.

mwcampbell:
The checked-in accesskit.h is now empty. BTW, I wonder if we actually need to check in that file. Does it simplify CI to do that?

DataTriny:
I checked it in simply because it's common practice. I guess it helps if people can see the headers without having to run any command while cloning the repo. I re-generate it before uploading to the release, if you want I can remove it entirely.

mwcampbell:
I think it would be better to not check in the header file. So then, I guess it shouldn't be stored under src, but only somewhere under target. If people want to look at the header file, they can look at the release artifact. And maybe it could also be built and uploaded as an artifact during CI, to make it easier to look at it when reviewing PRs.

mwcampbell:
I don't believe the build.rs script in bindings/c is actually necessary. We also shouldn't need the `links` option in Cargo.toml. That's only needed when building a binding in the other direction, i.e. a Rust binding for a C library.

DataTriny:
Yes you're right. I'll fix that and the CI tomorrow.

DataTriny:
Finally got the CI/CD working again. Please check out [here](https://github.com/DataTriny/accesskit/releases/tag/accesskit_c-v0.1.0) to see how releases will look like. Are there other major platforms we want to distribute binaries for? Maybe macOS/ARM? BTW I've chosen random names for the lib/* directories, we can change them.

mwcampbell:
We definitely need to build for macOS on ARM64. Fortunately, in my experience, cross-compiling for another macOS target on macOS is straightforward; just pass `--target`.

I would use `macos` rather than `darwin` in the macOS library directory names, for consistency.

The Windows library directories are missing the `accesskit.dll.lib` import library for the DLL.

DataTriny:
The new test release can be found [here](https://github.com/DataTriny/accesskit/releases/tag/accesskit_c-v0.1.0).

bruvzg:
> Are there other major platforms we want to distribute binaries for?

For the static libraries, it's probably worth having both `*-pc-windows-gnu` and `*-pc-windows-msvc` since these won't be compatible (DLLs and import libs should be cross-compatible).

DataTriny:
@bruvzg Yes, that's also what I had in mind. I'll consider it later.

mwcampbell:
The macos-arm64 build in the latest test release is broken. It is in fact another x86_64 build.

DataTriny:
A new test release is [here](https://github.com/DataTriny/accesskit/releases/tag/accesskit_c-v0.1.0).

Just by looking at the file size, the macOS ARM binaries are different now. The issue affected 32bit builds as well by the way.

I've added builds for MinGW. I think the CD part of this PR is done now, but let me know if you think something is still missing.

I still have a few types to rework but this PR should be ready tomorrow.

mwcampbell:
A few things about the prebuilt package:

* If we're going to leave the dynamic libraries on non-MSVC build targets unstripped, and I think we should, then I think we should also include accesskit.pdb for the Windows MSVC builds.
* The package should include the examples directory. This implies that the Windows example should be buildable from the package, not just from a repo working copy. Maybe we want to switch to storing the built header file in bindings/c/include in the repo, so the example can be built from either environment in the same way.

Otherwise, looks good.

DataTriny:
> Maybe we want to switch to storing the built header file in bindings/c/include in the repo

Do you want me to check back in `accesskit.h`?

mwcampbell:
I think it's still better not to check it in; the change I suggested would just mean adding an entry to bindings/c/.gitignore.

DataTriny:
There's only one big thing I don't know how to handle best. With the latest changes the Windows example doesn't build because libraries are in a different place in relation to `build.bat` depending on whether you build it from the repo or from the downloaded package. So I've included both paths in the `link` command, but then it complains that it can't find one or the other, and I can't find a simple way to just ignore that.

mwcampbell:
OK, I think it's more important to make the example build in the downloaded package. We can figure out how to make it build in the repository later. Also, can we add a win-arm64-msvc variant, using the aarch64-pc-windows-msvc target? Note though that there's no aarch64-pc-windows-gnu, so we can't have a win-arm64-mingw target.

DataTriny:
Oh actually there is another inconvenience on Windows. For now you'd have to manually create `bindings/c/include` and generate `accesskit.h` by running `cbindgen` yourself. Of course for other platforms there is a `Makefile`. I've seen other projects providing CMake files for Windows, should we do the same, or just leave instructions in the README?

Also, do you intend on publishing this crate so other bindings can depend on it? If yes, then I'll fill in the `Cargo.toml`.

mwcampbell:
I'm noticing something weird about the way you're pushing your commits. It looks like you're replaying your earlier commits with revisions. So I can't tell what you changed in your last push.

mwcampbell:
No, I don't intend to publish this crate. I don't see the point of any other Rust crate depending on it. We might even want to set `publish = false` in Cargo.toml so it's not published by accident.

DataTriny:
Sorry I was doing that to keep git history clean. I should have stopped doing that since I marked this ready for review. I just modified `bindings/c/examples/windows/build.bat` and `.github/workflows/bindings.yml` (to add the new Windows build).

mwcampbell:
I always squash the PR branch when I merge it, so the top-level history is clean.

DataTriny:
As a finishing touch, I wanted to make sure we never panic as it's undefined behavior across the FFI boundary, unfortunately this can't easily be done for the Windows and macOS adapters as types need to be [RefUnwindSafe](https://doc.rust-lang.org/std/panic/trait.RefUnwindSafe.html).

mwcampbell:
I don't like this approach to silently recovering from panics. In particular, a panic when applying a tree update may leave the adapter in an undefined state. And if a panic occurs, a C user won't know what went wrong; all information about the error will just be silently dropped. What if we used `panic = "abort"` instead, and just let the panic happen? When an unexpected error happens, I think it's better to just abort immediately, providing as much information as possible, and leaving the program in a state where the problem can be diagnosed using a debugger. `panic = "abort"` does that.

mwcampbell:
Sometimes it's easier to provide feedback by just doing fixes myself. So please look at my [c-bindings-fixes branch](https://github.com/AccessKit/accesskit/tree/c-bindings-fixes). In particular, I did several fixes in the Windows example.

DataTriny:
Lot of stupid mistakes from me, some I fixed yesterday and lost in an aborted git rebase...

I've integrated your changes, just removed a useless null pointer check in the example.

Setting `panic = "abort"` would still be UB apparently, and we'd have to apply it at the workspace level. The proper way might be by using `pub extern "C-unwind" fn ...`, but it's [not yet stabilized](https://github.com/rust-lang/rust/issues/74990). I guess the typical way to solve this would be to introduce global `get_last_error`/`clear_last_error` functions that the user would have to use in case of a function returning `NULL` or `FALSE`.

mwcampbell:
> Setting panic = "abort" would still be UB apparently, and we'd have to apply it at the workspace level.

I've been thinking about applying it at the workspace level anyway. But where did you read that it would be UB?

DataTriny:
If I understand the Rustonomicon well, it's UB if an exception comes from foreign code, so in our case if a callback fails. I guess there is not much we could do in that situation anyway, so I'll turn on aborting on panic.

mwcampbell:
I strongly believe that we shouldn't silently drop errors, e.g. by returning null or another default value if one of the arguments is null. It's much better to fail immediately, so the user knows exactly what is wrong and can easily debug it. Please look at my recent commit on the c-bindings-fixes branch, where I modified the bindings to panic on errors instead.

Also, I fixed a double-free of the UIA init marker in the Windows example.

If these fixes are OK with you, then after merging them, please do another test release, and if that goes well, I'm ready to merge this.

DataTriny:
I'll have a look tomorrow. I just remembered that I didn't fix `custom_actions` which can't easily be freed at the moment. I guess functions returning this struct should return a pointer to it instead.

DataTriny:
I've integrated your changes, but I still have a bit of work to do around arrays and slices, and I don't want to rush it. So no delivery for today unfortunately.

DataTriny:
@mwcampbell hopefully the final review please?

Test package: [accesskit_c-v0.1.0.zip](https://github.com/DataTriny/accesskit/releases/download/accesskit_c-v0.1.0/accesskit_c-v0.1.0.zip)

mwcampbell:
Looks good to me. I just need to take care of the release-please token now.

DataTriny:
Hi @bruvzg ,

I'll need a bit more time to understand the "install name" issue, as I'm not familiar with macOS at all.

As for the naming of the libraries, I was thinking that you would only keep the version that you actually use, and remove the other files. I'd recommend you use static libraries whenever possible.

mwcampbell:
I definitely see the usefulness of putting static and shared libraries in different directories. We want to make it easy to drop in the prebuilt package, and to drop in updates as they come out, without requiring the user to manually strip down the package.

Also, now that I think about it, we may want to switch to a deep directory hierarchy, for example: lib/windows/x86_64/msvc/shared

DataTriny:
@bruvzg `install_name` for macOS shared library was fixed in #247

PR #231, opened 2023-03-25
fix: Fix problems related to the root node

mwcampbell:
AccessKit allows the root node to omit the bounding rectangle. On Windows and Linux, we can use the window's own rectangle as a fallback. This PR implements the equivalent behavior on macOS.

### DIFF
--- a/platforms/macos/src/adapter.rs
+++ b/platforms/macos/src/adapter.rs
@@ -101,10 +101,7 @@ impl Adapter {
         let state = tree.state();
         let root = state.root();
         let point = from_ns_point(&view, &root, point);
-        if let Some(node) = root.node_at_point(point, &filter) {
-            return Id::autorelease_return(self.context.get_or_create_platform_node(node.id()))
-                as *mut _;
-        }
-        null_mut()
+        let node = root.node_at_point(point, &filter).unwrap_or(root);
+        Id::autorelease_return(self.context.get_or_create_platform_node(node.id())) as *mut _
     }
 }
--- a/platforms/macos/src/appkit/view.rs
+++ b/platforms/macos/src/appkit/view.rs
@@ -47,5 +47,11 @@ extern_methods!(
 
         #[sel(backingScaleFactor)]
         pub(crate) fn backing_scale_factor(&self) -> CGFloat;
+
+        // NSView actually implements the full NSAccessibility protocol,
+        // but since we don't have complete metadata for that, it's easier
+        // to just expose the accessibilityFrame method here.
+        #[sel(accessibilityFrame)]
+        pub(crate) fn accessibility_frame(&self) -> NSRect;
     }
 );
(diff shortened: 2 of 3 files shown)


mwcampbell:
The scope of this PR expanded somewhat, based on feedback from a user who was trying to examine AccessKit nodes with Accessibility Inspector.

mwcampbell:
@spencerudnick Can you please review this?

spencerudnick:
I didn't try building and running anything yet, but in principle these changes look correct. It's nice to get hit-testing working!

Let me know if you would like me to test things locally.

mwcampbell:
Merging this now based on the feedback I already got from the user who raised the problems that this PR fixes.

platforms/macos/src/adapter.rs
```diff
@@ -101,10 +101,7 @@ impl Adapter {
         let state = tree.state();
         let root = state.root();
         let point = from_ns_point(&view, &root, point);
-        if let Some(node) = root.node_at_point(point, &filter) {
```
artursapek, line 104: hi @mwcampbell - I've arrived at this change from `bevy`, which relies on `accesskit`. this specific code change adds significant latency to bevy registering mouse click events on my Mac, running `13.5 (22G74)`.

bevy `0.11` upgraded accesskit in https://github.com/bevyengine/bevy/pull/8655/files which included this change. there is now a noticeable delay in mouse click events registering on my machine.

I've recorded a video to show what I mean, you can hear when my mouse clicks and when the event registers in bevy. on the left is the updated version of your code, and on the right is the previous version.

https://imgur.com/a/py58Zk4

reverting this specific change in `platforms/macos/src/adapter.rs` fixes the latency.

I'm not familiar with this crate or code, but I'm guessing that `get_or_create_platform_node(node.id())` with `root` is slow? can you help me understand why this was necessary? can you think of a reason why it might be adding latency?

in the meantime I'm going to open an issue in the Bevy repo at their request.

thanks in advance for any help

artursapek: logged the issue: https://github.com/bevyengine/bevy/issues/9391

mwcampbell: You may need to post your video somewhere else, because on imgur, I don't hear anything.

Are you using any macOS accessibility features such as VoiceOver?

mwcampbell: Disregard my comment about Imgur; I found the volume control. But that video won't help me anyway. You said I could hear the difference, but apparently I need to be able to see as well in order to understand it. Anyway, I'll continue the discussion on the Bevy issue.

artursapek: oh are you blind? I'm sorry I didn't know that

basically it shows debug prints in the console happening much later than the audible mouse click

PR #239, opened 2023-04-01
fix: Improve C bindings package directory structure

DataTriny:
Addresses https://github.com/accesskit/accesskit/pull/230#issuecomment-1490852698, initially raised by @bruvzg

Test package: [accesskit_c-v0.2.0](https://github.com/DataTriny/accesskit/releases/download/accesskit_c-v0.2.0/accesskit_c-v0.2.0.zip)

I've used these workflow runs as an opportunity to try and fix the install name of the dynamic library on macOS. Looking at the logs I see a result, but does it actually work? If anyone with access to a Mac could confirm, that would be wonderful!

### DIFF
--- a/.github/workflows/bindings.yml
+++ b/.github/workflows/bindings.yml
@@ -12,34 +12,52 @@ jobs:
         include:
           - os: macOS-latest
             target: aarch64-apple-darwin
-            path: macos-arm64
+            path: macos/arm64
+            shared-ext: dylib
+            static-ext: a
           - os: macOS-latest
             target: x86_64-apple-darwin
-            path: macos-x86_64
+            path: macos/x86_64
+            shared-ext: dylib
+            static-ext: a
           - os: windows-2019
             target: aarch64-pc-windows-msvc
-            path: win-arm64-msvc
+            path: windows/arm64/msvc
+            shared-ext: dll
+            static-ext: lib
           - os: windows-2019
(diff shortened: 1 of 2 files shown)


mwcampbell:
The new directory structure looks good overall. And on macOS, based on the output of a command like:

```
otool -L lib/macos/arm64/shared/libaccesskit.dylib
```

I believe the install_name problem is fixed. But, since we don't yet have a sample that can run on macOS, I don't have an easy way of actually testing it.

I have the following feedback on Windows:

* The `build.bat` for the Windows example needs to be updated.
* In the msvc directories, accesskit.pdb needs to be moved into the shared subdirectory, since the .pdb file is only for the DLL.
* In the msvc/shared directories, accesskit.dll.lib should be renamed to accesskit.lib, since it's no longer in the same directory as a static library in the package; this will match what people normally expect for DLL import libraries. Likewise, though I'm a little less sure of this, I think we want to rename libaccesskit.dll.a to libaccesskit.a in the mingw/shared directories.

DataTriny:
[accesskit_c-v0.2.0.zip](https://github.com/DataTriny/accesskit/releases/download/accesskit_c-v0.2.0/accesskit_c-v0.2.0.zip)

The command you showed here is exactly the one I ran on the runner. The fix will come in another PR anyway.

mwcampbell:
The package looks good to me now.

PR #243, opened 2023-04-24
fix: Remove objc and Core Foundation types from C bindings public API

### LINKED ISSUE #242: Split platform specific "C"-bindings into the separate files
Currently, `accesskit.h` includes some Obj-C headers when it is used on macOS, therefore it is usable only from the Obj-C/Obj-C++ source, but not from pure C/C++.

Splitting platform specific adapter bindings to the separate file would allow using platform independent part of the code from C/C++.

Or alternatively, since no Obj-C specific stuff is really used in the binding, it might be worth getting rid of the Obj-C includes, by replacing `NSObject *` with `void *` (already done with `NSView *` for subclassing adapter) and `CGPoint` with two floats.

DataTriny:
Fixes #242 

It compiles, does it actually do what it is supposed to?

Feedback appreciated.

CC @bruvzg

### DIFF
--- a/bindings/c/src/macos.rs
+++ b/bindings/c/src/macos.rs
@@ -87,14 +87,15 @@ impl macos_adapter {
         adapter.focus() as *mut _
     }
 
+    /// Returns a pointer to an `NSObject`. Ownership of the pointer is not transfered.
     #[no_mangle]
     pub extern "C" fn accesskit_macos_adapter_hit_test(
         adapter: *const macos_adapter,
         x: f64,
         y: f64,
-    ) -> *mut NSObject {
+    ) -> *mut c_void {
         let adapter = ref_from_ptr(adapter);
-        adapter.hit_test(NSPoint::new(x, y))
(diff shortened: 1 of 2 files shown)


bindings/c/src/macos.rs
```diff
@@ -74,29 +71,30 @@ impl macos_adapter {
...
+        adapter.focus() as *mut _
     }
 
     #[no_mangle]
     pub extern "C" fn accesskit_macos_adapter_hit_test(
         adapter: *const macos_adapter,
-        point: NSPoint,
+        x: f64,
+        y: f64,
     ) -> *mut NSObject {
```
bruvzg, line 95:
```suggestion
    ) -> *mut c_void {
```

DataTriny: Woops, good catch! Sorry about that.

DataTriny:
Re-generated headers can be found here: https://github.com/AccessKit/accesskit/suites/12450808396/artifacts/662973745

bruvzg (approved):

PR #245, opened 2023-04-29
fix: don't remove ids which are to be added by a tree update

wash2:
I'm not sure if re-using IDs is possibly a usage error on my part, but I assume it is ok?

When I do reuse IDs, I found that using my current implementation with some Iced applications, a container widget like a scrollbar could change IDs between updates, while not all of its children do. Without this fix, the scrollbar's children are orphaned and removed despite still being children of the scrollbar, which just has a different ID.

### DIFF
--- a/consumer/src/tree.rs
+++ b/consumer/src/tree.rs
@@ -85,6 +85,13 @@ impl State {
             }
         }
 
+        // node ids tracks the ids of all nodes in the update.
+        // none of these nodes are orphans.
+        let node_ids = update
+            .nodes
+            .iter()
+            .map(|(id, _)| *id)
+            .collect::<HashSet<_>>();
         for (node_id, node_data) in update.nodes {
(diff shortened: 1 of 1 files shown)


DataTriny:
If a parent's ID changes, you are effectively creating a new node. Trying to add already existing nodes as children of this new node is invalid since they are still children of the old parent node. To remove the old parent, you have to include the old parent's parent in the same update, with its children not containing old parent's ID anymore.

Note however that creating new nodes that are effectively just clones of old nodes but with different IDs will raise extra accessibility events therefore leading to a probably bad user experience. The fundamental issue here is that your scrollbar should not change ID between updates, as long as you are referring to the same widget.

wash2:
Ok thanks, that means I'll probably have to generate the widget IDs a bit differently 😅 thanks for the clarification. I'll close this PR.

PR #251, opened 2023-06-19
chore: Add compiler settings to our release profile

mwcampbell:
I've wanted to add these for a while, and we'll soon release a new version of our C bindings, so now is a good time.

The `debug` flag enables as much debug info as possible in release builds without compromising optimizations. This is important for doing postmortem debugging of problems found in release builds. Windows uses a setting like this in its release builds, and I believe Chromium does too.

Maximizing optimization for size is perhaps more controversial, but I believe it's the best choice for us. Our library will look better to some decision-makers if it's smaller, and the difference is particularly dramatic with the Unix adapter. If we ever find that this decision noticeably hurts real-world performance, I'll be happy to reverse it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,3 +17,6 @@ default-members = [
 [profile.release]
 lto = true
 panic = "abort"
+debug = true
+opt-level = "z"
+codegen-units = 1


geom3trik (approved):

DataTriny:
Have you noticed significant compile times increase with `codegen-units = 1`? I know it makes it easier for the compiler to perform optimizations, but we also got complaints about the Unix adapter compile time.

mwcampbell:
No I haven't. And anyway, this only affects our own release builds, normally done in CI, not the builds of downstream Rust users in their own workspaces.

DataTriny (approved):
Fair enough!

DataTriny:
Removing `debug=true` on commit `424de4f` make Unix accesskit_c shared lib go from 20M to 3.9M. This is a very very high price that I think no user of our C bindings would be willing to pay.

PR #253, opened 2023-06-20
feat: Add window-based constructor to macOS subclassing adapter

mwcampbell:
This easily solves the issue we're having with #250, where we'd otherwise have to either write a little Objective-C or do some low-level stuff with the ObjC runtime from C code to make the SDL example compile on macOS. It might also help some Rust users, which is why I implemented the alternate constructor in the platform adapter itself.

### DIFF
--- a/platforms/macos/src/subclass.rs
+++ b/platforms/macos/src/subclass.rs
@@ -20,7 +20,11 @@ use objc2::{
 use once_cell::{sync::Lazy as SyncLazy, unsync::Lazy};
 use std::{collections::HashMap, ffi::c_void, sync::Mutex};
 
-use crate::{appkit::NSView, event::QueuedEvents, Adapter};
+use crate::{
+    appkit::{NSView, NSWindow},
+    event::QueuedEvents,
+    Adapter,
+};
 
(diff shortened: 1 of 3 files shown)


platforms/macos/src/subclass.rs
```diff
@@ -171,6 +184,25 @@ impl SubclassingAdapter {
         }
     }
 
+    /// Create an adapter that dynamically subclasses the content view
+    /// of the specified window. Panics if the specified window doesn't
```
DataTriny, line 188: I'd put the note about panics in a dedicated `# Panics` section, like the stdlib does.

DataTriny (commented):
Are you sure we won't need a similar function on the regular `Adapter`?

mwcampbell:
I've chosen not to add a similar function on the regular adapter because the assumption is that if a user is using the regular (non-subclassing) adapter, then they're implementing their own `NSView` subclass already, which means either they're writing in Objective-C, or they already have a way of working with the ObjC runtime, such as the objc2 crate. So they already have a way of directly providing the view, probably because they instantiated it from their own class.

DataTriny (approved):

PR #256, opened 2023-07-20
deps!: Bump winit to 0.29

Vrixyz:
I'm working on updating bevy to next winit version: https://github.com/bevyengine/bevy/pull/8745 ; a simple winit bump seems enough, but I noticed examples failing so I updated them. I'm sharing in case it's useful.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -52,7 +52,7 @@ version = "0.10.0"
 dependencies = [
  "accesskit",
  "accesskit_consumer",
- "objc2",
+ "objc2 0.3.0-beta.3.patch-leaks.2",
  "once_cell",
 ]
 
@@ -83,7 +83,7 @@ dependencies = [
  "scopeguard",
  "static_assertions",
  "windows",
- "winit",
+ "winit 0.28.3",
 ]
 
 [[package]]
@@ -94,7 +94,8 @@ dependencies = [
  "accesskit_macos",
  "accesskit_unix",
  "accesskit_windows",
- "winit",
+ "raw-window-handle 0.5.2",
+ "winit 0.29.3",
 ]
 
 [[package]]
@@ -113,6 +114,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "ahash"
+version = "0.8.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "91429305e9f0a25f6205c5b8e0d2db09e0708a7a6df0f42212bb56c32c8ac97a"
+dependencies = [
+ "cfg-if",
+ "getrandom",
+ "once_cell",
+ "version_check",
+ "zerocopy",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -128,15 +142,36 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c77a0045eda8b888c76ea473c2b0515ba6f471d318f8927c5c72240937035a6"
 dependencies = [
  "android-properties",
- "bitflags",
+ "bitflags 1.3.2",
  "cc",
  "jni-sys",
  "libc",
  "log",
- "ndk",
+ "ndk 0.7.0",
  "ndk-context",
- "ndk-sys",
- "num_enum",
+ "ndk-sys 0.4.1+23.1.7779620",
+ "num_enum 0.5.11",
+]
+
+[[package]]
+name = "android-activity"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "052ad56e336bcc615a214bffbeca6c181ee9550acec193f0327e0b103b033a4d"
+dependencies = [
+ "android-properties",
+ "bitflags 2.4.1",
+ "cc",
+ "cesu8",
+ "jni",
+ "jni-sys",
+ "libc",
+ "log",
+ "ndk 0.8.0",
+ "ndk-context",
+ "ndk-sys 0.5.0+25.2.9519653",
+ "num_enum 0.7.1",
+ "thiserror",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


mwcampbell:
I'm not sure I understand the point of this PR. Is this meant to prepare for winit 0.29? If so, I can't merge it until 0.29 is published. As a published crate, accesskit_winit can only depend on published versions of other crates; it can't depend on a git checkout of winit.

Vrixyz:
You're correct, it's not merge-able currently. The point is to prepare for a future winit 0.29, I should have put it in a draft status indeed. if you prefer a well finished PR feel free to ignore this and I can tag you again when it's merge-able (when winit releases, if you need it, if I get to it)

Cargo.lock
Vrixyz, line 1: This should have minimal changes, see https://github.com/AccessKit/accesskit/pull/257#issuecomment-1643725059

platforms/windows/src/tests/subclassed.rs
```diff
@@ -61,16 +60,26 @@ impl ActionHandler for NullActionHandler {
 // This module uses winit for the purpose of testing with a real third-party
 // window implementation that we don't control.
 
+#[cfg(target_os = "windows")]
```
Vrixyz, line 63: not sure that's the best way to handle that. I assumed it makes little sense to test other platforms in platform/windows 🤔

DataTriny: The `accesskit_windows` crate compiles just fine on Linux, and probably on macOS as well. Therefore one might try to run these tests outside of a Windows environment. So yes, this check makes sense to me.

Vrixyz:
I'll need it for https://github.com/bevyengine/bevy/pull/8745 ; so I'm planning to work on it if/when needed, but it's more a "side quest" for me. If anyone wants to work on it before I'm ok with it.

If waiting for me, I guess you should expect it to be ready in 1-2 months (not taking into account reviewing)

torokati44:
>  I want to update Vizia to use new version of winit, but Vizia also have accesskit as dependency

I'm in a similar situation with Ruffle, so if you were to finish this, I would highly appreciate it!

Vrixyz:
I updated it just so it compiles and unblocks me for further compiling my other work. I spotted a few rough edges I think I'll pass on on fixing,I'm just sharing my work if it can be useful:

- HasRawWindowHandle is deprecated but I still used it to avoid breaking everything.
- I added a feature rwh_06, not sure it's really needed.

Feel free to reuse/continue my PR or not, I think I won't work on it more, unless I spot blocking issues.

Vrixyz:
I needed to rely on rwh_05 (raw_windows_handle version 0.5) because the version of wgpu I use is still on rwh_05, ideally I guess accesskit could support all raw_window_handle versions by providing different features. I'll consider that out of the scope of that PR (which I consider focused on updating to winit 0.29, with minimal changes.

DataTriny:
Thank you @Vrixyz for your work.

Since interest seem high on this one, I'll push it to completion. I have everything fixed, just waiting for #301 to fix CI.

DataTriny:
OK, I think I'm done here.

DataTriny:
Of course! Waiting for a review from @mwcampbell. Please be patient.

tronical:
I've ported Slint to use this PR (we're using a newer winit version) and I can report that the changes work well for me (in https://github.com/slint-ui/slint/pull/3833 ).

DataTriny:
I think this PR was tested enough by now and I don't want to block everybody for too long. Unless I'm told otherwise, I'll merge this tomorrow.

I want to fix #300 before we release the next version, and possibly update raw-window-handle to 0.6 (feedback welcome on this one).

mwcampbell:
I'm finally reviewing this. Sorry for the delay.

mwcampbell:
I think we need to move the updates to the Windows adapter to a separate PR. This main PR has to be marked as a breaking change in the accesskit_winit crate. But the update to the accesskit_windows crate is not a breaking change, as it only affects test code. I know it's inconvenient, but given the way our semi-automated release process works, I think it's necessary.

DataTriny:
You're right, it would mess up accesskit_windows semver... I'll revert these changes.

DataTriny:
I'll need to rebase...

Vrixyz:
> I'll need to rebase...

I merged main into my branch before, so rebase might be a bit trickier than a merge (sorry about that) good luck 🤞

DataTriny:
> I merged main into my branch before, so rebase might be a bit trickier than a merge (sorry about that) good luck 🤞

Yeah, I squashed everything into a single commit because I really didn't feel like going through each one to fix conflicts. Sorry for erasing the git history as well as your name from these commits.

DataTriny:
Hmm, interesting... The tests are failing on our Windows pipeline, but they pass on my machine. Not much information other than a lock was poisoned.

mwcampbell:
I wonder if our recent change to use `Rc<ReceivedFocusEvent>` rather than `Arc<ReceivedFocusEvent>` was actually incorrect. I'm going to review that code and figure out if that pointer is actually sent across threads.

DataTriny:
Re-running the job didn't fix the issue so at least it's reproduceable. I still can't make it fail on my machine, even if I run the tests from Powershell.

mwcampbell:
Yes, it appears my guess was right. Because I deliberately decided to create the UIA client on a thread where COM is initialized in MTA mode, any UIA event handlers are called on a separate thread, and using `Rc` in such event handlers is unsafe. I'll have to fix this in a separate PR.

mwcampbell:
I couldn't make it fail either, on my quad-core laptop. As with many concurrency bugs, it's probably only intermittently reproducible, and the conditions happen to be right in the GitHub Actions runner at the moment.

DataTriny:
Well... At least now we don't panic, but we still have one failing test:

```
thread 'tests::simple::focus' panicked at platforms\windows\src\tests\simple.rs:189:9:
assertion failed: equal
```

mwcampbell:
Why change the macOS backend, and only the macOS backend, to use `crate::WindowEvent` rather than importing `WindowEvent` from winit?

mwcampbell:
I don't know what to do about the failing test. It passed in the CI run for my PR about half an hour ago. Maybe a VM is being reused for the CI runs for this PR, and it's in a weird state. The test also passes for me locally on this branch.

DataTriny:
No idea, I don't remember making this change as I can't test it anyway. I see that I forgot to remove a comment on this file as well.

DataTriny:
I was able to get the test failing once on my machine through the newer Terminal application. It occurred the first time I tried running the tests from this app and all subsequent runs passed.

mwcampbell:
OK, this PR looks good to me now. And now the Windows tests passed. Go figure. May I merge?

DataTriny:
I went through all the changes once again and I think we are good!

These spurious test failures probably have nothing to do with this PR, so let's merge this.

mwcampbell:
For anyone who missed it, I published a new version of accesskit_winit yesterday. So if you've been waiting on AccessKit to update to winit 0.29 before you do the same, you're now unblocked.

PR #263, opened 2023-07-30
refactor!: New C API for tree updates

mwcampbell:
`accesskit_tree_update` is now an opaque struct like the other wrapper structs around boxed types. I think this API is safer, lower-overhead, and more pleasant to use (as C APIs go). It will also let us experiment with passing a serialized tree update across the FFI boundary, as we did in the proof-of-concept Unity plugin, by adding a function like `accesskit_tree_update_from_json`.

### DIFF
--- a/bindings/c/src/common.rs
+++ b/bindings/c/src/common.rs
@@ -911,8 +911,6 @@ impl tree {
     }
 }
 
-opt_struct! { opt_tree, tree }
-
 impl From<tree> for Tree {
     fn from(tree: tree) -> Self {
         Self {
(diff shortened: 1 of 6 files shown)


bindings/c/src/common.rs
```diff
@@ -922,65 +920,72 @@ impl From<tree> for Tree {
...
+        id: node_id,
+        node: *mut node,
+    ) {
+        let update = mut_from_ptr(update);
+        let node = box_from_ptr(node);
+        update.nodes.push((id.into(), *node));
+    }
+
+    #[no_mangle]
+    pub extern "C" fn accesskit_tree_update_set_tree(update: *mut tree_update, tree: tree) {
```
DataTriny, line 968: For consistency with other structs, I would have kept the `opt_tree` struct and added an `accesskit_tree_update_get_tree` method. I know this struct is mostly meant to be written into, and not read, but so is `NodeBuilder`, yet we provide getters.

bindings/c/src/common.rs
```diff
@@ -922,65 +920,72 @@ impl From<tree> for Tree {
...
+    }
+
+    #[no_mangle]
+    pub extern "C" fn accesskit_tree_update_clear_tree(update: *mut tree_update) {
+        let update = mut_from_ptr(update);
+        update.tree = None;
+    }
+
+    #[no_mangle]
+    pub extern "C" fn accesskit_tree_update_set_focus(update: *mut tree_update, focus: node_id) {
```
DataTriny, line 980: I would have created `accesskit_tree_update_get_focus` as well.

DataTriny (commented):
It's indeed easier and safer to use. It's just unfortunate the nodes now can't be read back.

mwcampbell:
In practice, `TreeUpdate` is only ever read by the consumer crate. Even if we eventually work on embedded ATs for environments with no accessibility APIs, those would still go through the consumer crate. And implementing such things in non-Rust languages isn't currently a priority for me. So I'm fine with making the tree update C API write-only.

The reason why I have separate set/clear methods rather than having the setters take `opt_tree` or `opt_node_id` is the same reason why I define set/clear methods on `NodeBuilder` even in Rust. I want to minimize the level of dynamism that's forced by the API, by having separate functions for the "some" and "none" cases. This keeps code size down in the `NodeBuilder` methods, because of how the dynamic property system works. And in the C API, it means that the setters don't have to convert from the `opt_*` struct to a Rust option. I hope that makes sense.

mwcampbell:
To clarify my previous comment, even when we do have bindings for the consumer crate, we still won't need the ability to read things back from the tree update in non-Rust code. So the fact that I'm not prioritizing language bindings for the consumer crate is irrelevant to this PR.

DataTriny:
It's fine with me. My comment was more on API consistency. I'm not a big fan of all the `opt_*` struct BTW, I just couldn't come up with something better...

DataTriny (approved):

PR #264, opened 2023-08-02
feat: Support the enumn crate in all public enums

mwcampbell:
I think I need this for my JNI bindings.


DataTriny:
@mwcampbell You forgot to include the changes to `Cargo.toml` and `Cargo.lock`.

mwcampbell:
It turns out that no such changes are needed, because the accesskit crate already has an optional dependency on enumn (and therefore, an enumn feature); that dependency/feature was needed for our serialization implementation.

DataTriny (approved):

DataTriny:
Indeed you're right, I missed that.

mwcampbell:
Merged as a fix rather than a feature because it's not a breaking change, and I don't want to bump semver-major version numbers for something this small.

PR #266, opened 2023-08-07
feat: Workaround for libraries that put the macOS keyboard focus on the window rather than the content view

mwcampbell:
This fixes the macOS focus problem discussed in #250.

@madsmtm I had to copy some code from objc2 in order to add a method to an existing class. That was the only solution I could find; creating a subclass and changing the class of the window instance led to very broken behavior (the window never appeared), likely because `NSWindow` uses KVO.

### DIFF
--- a/platforms/macos/src/patch.rs
+++ b/platforms/macos/src/patch.rs
@@ -1,4 +1,4 @@
-// Copyright 2022 The AccessKit Authors. All rights reserved.
+// Copyright 2023 The AccessKit Authors. All rights reserved.
 // Licensed under the Apache License, Version 2.0 (found in
 // the LICENSE-APACHE file) or the MIT license (found in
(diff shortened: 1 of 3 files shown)


platforms/macos/src/patch.rs
```diff
@@ -0,0 +1,90 @@
+// Copyright 2022 The AccessKit Authors. All rights reserved.
```
DataTriny, line 1:
```suggestion
// Copyright 2023 The AccessKit Authors. All rights reserved.
```

DataTriny (changes requested):
Probably better to let macOS experts have a close look at this one!

DataTriny (approved):

mwcampbell:
Mads said it was fine, so I've gone ahead and merged it.

PR #269, opened 2023-08-11
feat: Add Python bindings

DataTriny:
Adds a new crate containing bindings for the Python programming language.

Wheels will be automatically uploaded to `pypi.org` for the most common platforms/architectures.

The pygame example is on par with the SDL example, except for Linux where hit-testing is probably not possible due to pygame not exposing window real size.

I'm not 100% sure about the unsendable types: is it enough to just document the fact that they must only be used from the main thread? Or should we protect users more?

#### TODO

- [ ] Create a `PYPI_TOKEN` repository secret

### DIFF
--- a/.github/workflows/python-bindings.yml
+++ b/.github/workflows/python-bindings.yml
@@ -5,72 +5,139 @@ on:
 
 name: Publish Python bindings
 
-jobs:
-  build-wheels:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        include:
-          - os: macos-latest
-            python-arch: x64
-            rust-target: x86_64
-          - os: macos-latest
-            python-arch: x64
-            rust-target: universal2-apple-darwin
-          - os: ubuntu-latest
-            python-arch: x64
-            rust-target: x86_64
-          - os: ubuntu-latest
-            python-arch: x64
-            rust-target: i686
-            skip-wheel-installation: true
-          - os: windows-latest
-            python-arch: x64
-            rust-target: x64
-          - os: windows-latest
-            python-arch: x86
-            rust-target: x86
+env:
+  MIN_PYTHON_VERSION: 3.9
 
+jobs:
+  macos-wheels:
+    runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        target: [x86_64, aarch64]
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         with:
-          python-version: 3.7
-          architecture: ${{ matrix.python-arch }}
-      - uses: dtolnay/rust-toolchain@stable
-      - name: Build wheel
+          python-version: ${{ env.MIN_PYTHON_VERSION }}
+      - name: Build wheels
         uses: PyO3/maturin-action@v1
         with:
-          target: ${{ matrix.rust-target }}
-          manylinux: auto
-          args: --release --out dist --sdist
+          target: ${{ matrix.target }}
+          args: --release --out dist
+          sccache: 'true'
       - name: Test wheel installation
-        if: matrix.skip-wheel-installation != true
+        if: matrix.target == 'x86_64'
         run: |
           pip install accesskit --no-index --find-links dist --force-reinstall
           python -c "import accesskit"
-      - name: Upload wheel
+      - name: Upload wheels
         uses: actions/upload-artifact@v3
         with:
           name: wheels
           path: dist
 
-  release:
-    name: Release
+  unix-wheels:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        target: [x86_64, x86, aarch64]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.MIN_PYTHON_VERSION }}
+      - name: Build wheels
+        uses: PyO3/maturin-action@v1
+        with:
+          target: ${{ matrix.target }}
+          args: --release --out dist
+          sccache: 'true'
+          manylinux: auto
+      - name: Test wheel installation
+        if: matrix.target == 'x86_64'
+        run: |
+          pip install accesskit --no-index --find-links dist --force-reinstall
+          python -c "import accesskit"
+      - name: Upload wheels
+        uses: actions/upload-artifact@v3
+        with:
+          name: wheels
+          path: dist
+
+  windows-wheels:
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        target: [x64, x86]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.MIN_PYTHON_VERSION }}
+          architecture: ${{ matrix.target }}
+      - name: Build wheels
+        uses: PyO3/maturin-action@v1
+        with:
+          target: ${{ matrix.target }}
+          args: --release --out dist
+          sccache: 'true'
+      - name: Test wheel installation
+        run: |
+          pip install accesskit --no-index --find-links dist --force-reinstall
+          python -c "import accesskit"
+      - name: Upload wheels
+        uses: actions/upload-artifact@v3
+        with:
+          name: wheels
+          path: dist
+
+  sdist:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build sdist
+        uses: PyO3/maturin-action@v1
+        with:
+          command: sdist
+          args: --out dist
+      - name: Upload sdist
+        uses: actions/upload-artifact@v3
+        with:
+          name: wheels
+          path: dist
+
+  pypi-release:
+    name: Publish to PyPI
     if: "startsWith(github.ref, 'refs/tags/')"
-    needs: [build-wheels]
+    needs: [macos-wheels, unix-wheels, windows-wheels, sdist]
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/download-artifact@v3
         with:
           name: wheels
-      - uses: actions/setup-python@v4
-        with:
-          python-version: 3.9
-      - name: Publish to PyPI
+      - uses: PyO3/maturin-action@v1
         env:
-          TWINE_USERNAME: __token__
-          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
-        run: |
-          pip install --upgrade twine
-          twine upload --skip-existing *
+          MATURIN_PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
+        with:
+          command: upload
+          args: --non-interactive --skip-existing *
+
+  github-release:
+    name: Add to GitHub release
+    if: "startsWith(github.ref, 'refs/tags/')"
+    needs: [macos-wheels, unix-wheels, windows-wheels, sdist]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/download-artifact@v3
+        with:
+          name: wheels
+          path: dist
+
+      - uses: AButler/upload-release-assets@v2.0
+        with:
+          files: "dist/*"
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          release-tag: ${{ github.ref_name }}
--- a/.github/workflows/python-bindings.yml
+++ b/.github/workflows/python-bindings.yml
@@ -111,6 +111,9 @@ jobs:
 
   pypi-release:
     name: Publish to PyPI
+    environment: release
+    permissions:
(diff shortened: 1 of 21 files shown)


mwcampbell:
Can we configure pyo3 to rename the enum variants so they follow the ALL_CAPS_SNAKE_CASE style that's typical of Python, rather than the CamelCase style of Rust?

mwcampbell:
I suggest we format the example using [Black](https://black.readthedocs.io/en/stable/).

DataTriny:
> Can we configure pyo3 to rename the enum variants so they follow the ALL_CAPS_SNAKE_CASE style that's typical of Python, rather than the CamelCase style of Rust?

There is no automatic way of achieving this unfortunately. We'd have to add `#[pyo3(name = "...")]` to every enum variants.

I thought it was not a good idea to create new enums for this single purpose.

mwcampbell:
I really don't want to ship the Python bindings with that obviously unidiomatic style. That's the kind of thing that would be really tedious to fix later, particularly for users. I suggest implementing a derive macro to automatically add the PyO3-specific attribute on all the enum variants. This means adding a crate like `accesskit_macros` (could be in a top-level macros directory). Such a crate should only be in the dependencies of the base accesskit crate if the pyo3 feature is enabled.

mwcampbell:
Can we remove Black from the requirements.txt file that we provide with the example? Users don't need Black in order to run the example.

DataTriny:
> I suggest implementing a derive macro to automatically add the PyO3-specific attribute on all the enum variants.

After some testing, I'm not even sure if it's possible. If I add `#[pyo3]` attributes from a proc macro, the common crate refuses to compile because it doesn't understand the link between the `#[pyo3]` attribute and the `#[pyclass]` attribute.

Manually adding the full attribute that would be needed, that is `#[cfg_attr(feature = "pyo3", pyo3(name = "UNKNOWN"))]` to the first variant of the `Role` enum, triggers exactly the same error:

```
error: cannot find attribute `pyo3` in this scope
  --> common/src/lib.rs:57:34
   |
57 |     #[cfg_attr(feature = "pyo3", pyo3(name = "UNKNOWN"))]
   |                                  ^^^^
   |
   = note: `pyo3` is in scope, but it is a crate, not an attribute
```

Unless I miss something, implementing this directly inside PyO3 might be the only solution.

DataTriny:
I've opened pyo3/pyo3#3383 in that direction. But if we want to quickly ship something, duplicating the enums might be the only option.

mwcampbell:
I'm not in a hurry to release the Python bindings. Let's take the time to do it right. So I think we should wait for the response to the new PyO3 issue and see if either we or the PyO3 developers can implement something that helps all PyO3 users with this problem.

DataTriny:
My work to support renaming enum variants was merged into PyO3 so I'm switching to temporarily pull the master branch. There is discussion to release PyO3 0.20 in a few weeks.

As always, if someone can test the pygame example on macOS and report back here that would be nice. I would be surprised if it doesn't work out of the box since it's SDL under the hood.

mwcampbell:
Unfortunately, the Pygame example does not run on macOS. Attempting to instantiate the macOS adapter fails with the following error:

```
TypeError: argument 'window': 'PyCapsule' object cannot be interpreted as an integer
```

DataTriny:
Thanks @mwcampbell for the feedback.

Looking at pygame's source code it's now obvious. It's actually the safest way to pass a pointer, but I'm not sure we should only expect to receive a PyCapsule here. I'll see how other windowing libraries do this.

DataTriny:
While rebasing this I played a bit with ruff this morning. My understanding is that it doesn't do formatting but linting. As such it would not replace black. It was able to point me to a couple of issues in the pygame example, the question is: do we want to run these two tools just for one Python file? This would mean adding another step in the formatting CI job.

DataTriny:
@mwcampbell The Pygame example should work on macOS hopefully now. Can you please confirm?

leonardder:
> While rebasing this I played a bit with ruff this morning. My understanding is that it doesn't do formatting but linting. As such it would not replace black. It was able to point me to a couple of issues in the pygame example, the question is: do we want to run these two tools just for one Python file? This would mean adding another step in the formatting CI job.

Ruff has an experimental formatter that should be pretty similar to Black in its output. While it is not yet stable, given this project doesn't have that much Python code, I think it should be sufficient.

.github/workflows/python-bindings.yml
```diff
@@ -0,0 +1,76 @@
...
+            rust-target: x64
+          - os: windows-latest
+            python-arch: x86
+            rust-target: x86
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.7
```
simlay, line 38: [Python 3.7 was EOL'd  on `2023-06-07`](https://devguide.python.org/versions/). Also, this should probably match the python version in the `release` job which could be part of the matrix.

.github/workflows/python-bindings.yml
```diff
@@ -0,0 +1,76 @@
...
+          architecture: ${{ matrix.python-arch }}
+      - uses: dtolnay/rust-toolchain@stable
+      - name: Build wheel
+        uses: PyO3/maturin-action@v1
+        with:
+          target: ${{ matrix.rust-target }}
+          manylinux: auto
+          args: --release --out dist --sdist
```
simlay, line 46: It's my understanding (I could be wrong) that this will only build the wheel for the python version specified in the `actions/python@v4` action (in this case `3.7`). Getting a good matrix can be rough. In the past, [I've used `pypa/cibuildwheel` as seen here](https://github.com/infinyon/fluvio-client-python/blob/7d10bdb7c55ff720afcbd100050ca5d4d0198ffb/.github/workflows/publish.yml#L9-L71) which definitely has some negatives (the whole build runs in a docker container which can have different libraries than the host).

simlay (commented):
Hope you don't mind a random drive by review, I saw this and felt like I might have some useful words. Feel free to disregard my comments if they're outside the goals of this PR. It's cool to see more Rust-Python wheels!

mwcampbell:
Thanks @simlay for the review. We certainly need more feedback from people like you who have previous experience building Python packages with Rust.

.github/workflows/python-bindings.yml
DataTriny: It was OK when I started this PR, but it should be updated indeed.

.github/workflows/python-bindings.yml
DataTriny: You are right. In fact, here is the link to the download page of a test upload I did: https://test.pypi.org/project/accesskit/#files

simlay: TIL about test.pypi.org. Cool.

Well, I don't work on the Python client for Fluvio anymore (and it looks like there hasn't been a release in a while) but it's my recollection that `pypa/cibuildwheel` action took a long time (~30 minutes) because each of the python versions was done in serial.

One thing that I also recall being a pretty rough time was if there's OpenSSL 3 in the dependency tree. This uses a different version of perl than OpenSSL 1 and the pypa images didn't have it. I'm currently failing at finding the github issue for this :/

You might be able to get away with just adding a `python-version` list to your matrix and have them run in parallel. It's unclear what kind of target OS/arch/python version support you're looking for. [`pypa/cibuildwheel` has a lot of python versions and targets](https://github.com/pypa/cibuildwheel#what-does-it-do).

mwcampbell: No, zbus doesn't depend on OpenSSL. On Linux, we don't depend on any C libraries other than the base system libraries (libc, pthread, etc.)

DataTriny: I wasn't 100% sure as I am not familiar with D-Bus authentication mechanism, but no OpenSSL involved here thankfully!

mwcampbell:
Finally verified that the pygame example runs on macOS. Reviewing the code now.

mwcampbell:
I'm not sure I understand why it's necessary to place pyproject.toml at the root of the repository, rather than under bindings/python. Is it just so we can include the license files in the built packages, or is there another reason as well?

mwcampbell:
Oh, I think I understand why pyproject.toml needs to be at the root of the repo. When you create a source distribution with `maturin sdist`, it will create the tarball from the root of the workspace and put pyproject.toml there in the root. So to ensure consistent behavior regardless of how the wheels are built, it's best to put pyproject.toml in the root of the repo. That also provides a convenient way to include the license files in the source distribution.

The problem is that when pyproject.toml is in the root of the repo, maturin runs `cargo build` from that root directory, which means that `bindings/python/.cargo/config` isn't read. This doesn't seem to actually be a problem for the macOS linker arguments, since maturin adds those anyway. But it means that we can't use `bindings/python/.cargo/config` to override the release profile, which I think we should do, to set `debug = false` and `panic = "unwind"` when building the Python binding, without affecting the C binding. So to get those release profile settings for the Python binding without moving away from a monorepo with a single workspace, I guess we need to submit a patch to maturin to run the build from the directory containing the Cargo manifest.

mwcampbell:
I think we should remove the `tools.maturin.exclude` entry from `pyproject.toml`. By default, maturin includes the examples directory in the source distribution, but not in the wheels, and I think that's what we want.

DataTriny:
> So to get those release profile settings for the Python binding without moving away from a monorepo with a single workspace, I guess we need to submit a patch to maturin to run the build from the directory containing the Cargo manifest.

Now I remember having troubles with this months ago but I didn't have a Mac to check. We can still manually pass the flags when we invoke maturin. That would be fine in the CI but something to remember when building locally.

> By default, maturin includes the examples directory in the source distribution, but not in the wheels...

Good catch, but I wonder why did I put this here then?

.github/workflows/python-bindings.yml
DataTriny: Coming back to this: we actually only need to build wheels for the oldest Python version we want to support because I'm using the limited ABI3 feature which provide backward compatibility. Since we are not doing much from the Python side, it's best to enable it and only build for, let's say, Pyton 3.9.

mwcampbell:
I think the changes to the common crate need to be moved to their own PR, with the `fix` prefix in its title, so as not to trigger a semver-breaking release of the common crate.

mwcampbell:
nit: In common.rs, I would rename `convertion_getter` and `convertion_setter` to `converting_getter` and `converting_setter`. I'm pretty sure "convertion" is a misspelling, and to me it makes sense to apply "converting" as an adjective here.

mwcampbell:
nit: In the Python wrapper for `Tree`, only the `toolkit_version` field isn't `pub`.

mwcampbell:
It looks to me like the Python binding for `ActionData` isn't actually usable in its current form. Remember that, in actual usage, `ActionData` is always passed from Rust to Python, so we don't need to be able to construct an `ActionData` from Python, but we do need to destructure it, that is, find out which variant we got and pull out the data inside.

Also, the `ActionData` and `ActionRequest` classes aren't properly registered in the Python module.

mwcampbell:
In the Python binding for the macOS `Adapter` class, I suggest that we wrap the results of the `view_children`, `focus`, and `hit_test` methods in a PyCapsule, or if possible, in a ctypes `c_void_p` pointer type. Given that there are multiple Python bindings for ObjC, I don't know what else to do with these Objective-C object pointers.

mwcampbell:
The doc comment for the Python Windows `SubclassingAdapter` class mentions the main thread. I suppose this was copied from the macOS version, but it's not applicable here.

There are still two references to Python 3.7: in the new CI workflow, and in the README for the pygame example.

My review is now complete.

DataTriny:
@mwcampbell I believe to have addressed all your comments.

I needed to test the CI on my fork before pushing it. I've attached the wheels to the GitHub release as well.

As for the macOS adapter, I went with a PyCapsule as it would be more cumbersome to create a `c_void_p` from Rust. I still had to cast the pointer though...

DataTriny:
I'd like to have #324 merged before we release accesskit_python 0.1 as the API of the Unix adapter will have to be updated.

mwcampbell:
The latest changes look good to me. So now I just have to figure out how we'll publish on PyPI.

DataTriny:
@mwcampbell I currently don't have access to a Linux desktop environment to test the pygame example. If you can do it before merging, otherwise I'll take care of it later. Thanks.

DataTriny:
The pygame example still works on Linux!

mwcampbell:
Just one more question before we merge this: It looks like the CI workflow doesn't currently use PyPI's "trusted publisher" feature. How hard would it be to use that feature?

DataTriny:
I've setup the workflow to use this feature. You'll need to configure it on the PyPI side as described [on this page](https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/).

Information you will have to provide as shown in a screenshot on the page:

- Repository name: accesskit
- Workflow name: python-bindings.yml
- Environment name: release

Not sure about switching from your personal account to an organization though.

PR #272, opened 2023-08-12
docs: Add CONTRIBUTING.md

### LINKED ISSUE #268: Add a CONTRIBUTING.md ?
Just a minor suggestion. A CONTRIBUTING.md or something along those lines might be helpful after running into these things that were non-obvious to me working on #257:
- Why we have a Cargo.lock file
- When should the Cargo.lock file be updated
- Changelog should not be modified. Instead use ?? (something to do with release-please, I don't understand it yet)
- How to run tests locally
   - run `cargo test -p accesskit_windows` in addition to `cargo test`
   - Note that some tests may fail if the window looses focus. Try running directly from terminal instead of from your IDEs terminal
   
If desired, I could make an initial draft based on these points, but if someone that works on the project regularly has time, they might have better knowledge of what should be in it.

paul-hansen:
Closes #268

"Edits by maintainers" is enabled and more than welcome!

This is my first time making a CONTRIBUTING.md so feel free to just use it as a starting point and/or go a completely different direction.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,12 @@
 # Contributing to AccessKit
 
+## Reporting an issue
+
+When reporting an issue, in order to help the maintainers understand what the problem is, please make your description of the issue as detailed as possible:
+
+- if it is a bug, please provide clear explanation of what happens, what should happen, and how to reproduce the issue, ideally by providing a minimal program exhibiting the problem
+- if it is a feature request, please provide a clear argumentation about why you believe this feature should be supported by AccessKit
+
 ## Making a Pull Request
 
 ### Cargo.lock File
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,15 +11,11 @@ When reporting an issue, in order to help the maintainers understand what the pr
 
 ### Cargo.lock File
 
-AccessKit intentionally includes the `Cargo.lock` file in our git repository.
-This is mainly due to our bindings to other languages.
+AccessKit intentionally includes the `Cargo.lock` file in the git repository.
 
-Usually you should not run cargo update when creating a pull request even when adding/changing a dependency.
-Simply building the library should update the `Cargo.lock` file with the minimal changes needed.
-
-> [!NOTE]
-> This is not normal / best practice for most libraries.
-> See the [official documentation](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) for more information. 
+You should not run `cargo update` when creating a pull request even when adding or changing a dependency.
+Simply building the library will update the `Cargo.lock` file with the minimal changes needed.
+Remember to commit these changes as part of your pull request.
 
 ### CHANGELOG.md
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,8 +12,6 @@ Simply building the library should update the `Cargo.lock` file with the minimal
 
 > [!NOTE]
 > This is not normal / best practice for most libraries.
-> If you are writing your own library, your `Cargo.lock` file is not used when the library is included in another project.
-> This means your CI will be testing with an out of date set of dependencies compared to your users, missing potential problems.
 > See the [official documentation](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) for more information. 
 
 ### CHANGELOG.md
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
+# Contributing to AccessKit
+
+## Making a Pull Request
+
+### Cargo.lock File
+
+AccessKit intentionally includes the `Cargo.lock` file in our git repository.
+This is mainly due to our bindings to other languages.
```
paul-hansen, line 8: A better defense of why this is needed might be a good addition here. I don't know the reason myself so this was the best I had.

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
+# Contributing to AccessKit
+
```
paul-hansen, line 2: A welcoming message, perhaps including something about what types of contributions you are looking for / open to would be fitting to have here. 

Felt like it would be disingenuous if I tried to write one as I'm not a regular maintainer.

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+```
+ERROR: Couldn't generate bindings for /home/runner/work/accesskit/accesskit/bindings/c.
+```
+Somewhere in the big block of text in the error message it also says:
+```
+unknown feature `proc_macro_span_shrink`
+```
+
+This is likely not an issue with your PR and just means `proc-macro2` needs to be updated in our Cargo.lock file. 
+You probably want to make this change in a separate PR as `proc-macro2` can update often and your PR could become out of date if it takes a while to be reviewed.
```
paul-hansen, line 71: Just lmk if you think this isn't a good place for this, was on the fence about including this myself. Not sure how often it actually happens etc.

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+#### Example Commit Messages
+Taken from [Conventional Commits documentation](https://www.conventionalcommits.org/en/v1.0.0/#summary):
+> feat: allow provided config object to extend other configs
+>
+> BREAKING CHANGE: `extends` key in config file is now used for extending other config files
+
+> fix: prevent racing of requests
+>
+> Introduce a request id and a reference to latest request. Dismiss
+incoming responses other than from latest request.
```
paul-hansen, line 34: I haven't used Release Please before. You might want different examples based on what you want to encourage use of.

Can also just remove the examples if you think the link to Conventional Commits is enough, I figured it might be nice to not have to leave the page to get a basic idea of it was my reasoning.

CONTRIBUTING.md
waywardmonkeys: I don't think that exact breakage would repeat. It would perhaps be something different (or perhaps not).

paul-hansen: Should I just take this section out, or do you think it's worth having, just need to make it not specific to the `proc_macro_span_shrink` feature? 

Could also take it out for now and consider adding something if it keeps being a problem in the future.

paul-hansen: Removed that section for now to keep things moving. Might be better in a future PR if it keeps being a problem. (or feel free to suggest a revision of it)

DataTriny:
Thank you for starting this!

I think we want to link to this page from the main README. I would add a Contributing chapter above the License one.

Something like this should be enough:

> Contributions to AccessKit are welcome. Please see CONTRIBUTING.md.

CONTRIBUTING.md
```diff
@@ -0,0 +1,53 @@
...
+
+> [!NOTE]
+> This is not normal / best practice for most libraries.
+> See the [official documentation](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) for more information. 
+
+### CHANGELOG.md
+
+Our CHANGELOG.md files are auto generated using [Release Please](https://github.com/googleapis/release-please) and should not be edited manually.
+
+To control what is in the CHANGELOG.md for your change, use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) when writing your commit messages.
```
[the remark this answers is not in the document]
DataTriny: We perform squash merges everytime and in this case it's the PR title that matters.

DataTriny:
@paul-hansen I've updated the file, trying to be exhaustive but concise. Let me know what you think about it.

paul-hansen:
Looks great!

I still do wonder if versioning the Cargo.lock file is a good decision, but we are documenting the current process so I agree with removing stuff trying to explain the reasoning from this file 👍

DataTriny (approved):

PR #273, opened 2023-08-16
Suppress non-snake-case warnings.

waywardmonkeys:


DataTriny:
Thank you @waywardmonkeys ,
Any guess on why these aren't picked up by clippy?
We allow `non_upper_case_globals` on entire files, should we do the same for these?

mwcampbell:
I assume VSCode is stricter about this, particularly for trait `impl`s, than clippy. @DataTriny Do you have any problem with this fix?

DataTriny:
I don't have an issue with the changes proposed here. I'm just wondering whether we should turn on more clippy checks.

DataTriny (approved):

PR #276, opened 2023-08-18
refactor!: Switch to simple unsigned 64-bit integer for node IDs

### LINKED ISSUE #275: RFC: Switch `NodeId` from `NonZeroU128` to `u64`
This is currently a request for comments; I haven't made a final decision to do this. But I'm opening it as an issue because, if we move forward with it, it will be actionable. And I want to make sure this gets resolved one way or the other.

I'm thinking of switching the underlying type for AccessKit node IDs from `NonZeroU128` to `u64`. The primary reason is that most type systems, for both programming languages and statically typed serialization formats, don't have a native type for an unsigned 128-bit integer. The need to have a struct for a node ID is one thing that makes the C API cumbersome, and the same concern applies to the work-in-progress Java API.

The original reason for switching from `NonZeroU64` to `NonZeroU128` last year was to allow a node ID to be constructed from an unsigned 64-bit integer that might be zero, such as a Rust hash code, without any risk of either losing data or breaking the "non-zero" requirement. This is an important concern, which is why I didn't hesitate to make the switch. But I've since reconsidered whether we really need to require that the node ID be non-zero. The original reason for that requirement was to optimize the size of `Option<NodeId>` in Rust. But with the node structure refactor that we introduced earlier this year, we no longer need to store many `Option<NodeId>` fields. So now I'm OK with using a plain `u64`.

My only concern at this point is that this change will interfere with the way slint's AccessKit integration currently composes a node ID from a component ID plus an index of an item within the component. The item index is currently a `usize`, which is 64-bit on modern general-purpose machines, meaning the node ID has to be more than 64-bit. One option would be to specify that the item index must be a `u32`, as it already is on some machines that slint targets (specifically, microcontrollers). Another solution would be for slint's AccessKit integration to maintain a complete mapping between items and node IDs. A final option I'm considering is some kind of subtree support, where each subtree has a node ID within the parent tree, and each subtree has its own node ID space. The problem that immediately comes to mind with that solution is that, with the current schema, there wouldn't be a way for a node in one subtree to refer to a node in a different subtree, such as when a control and its label are in different subtrees.

What we have right now works, but it's arguably a wart in the current design, particularly when considering FFI and serialization. We need to address all known warts as we move toward stabilization.

cc @raphlinus @wtholliday @tronical

mwcampbell:
fixes #275

### DIFF
--- a/consumer/src/lib.rs
+++ b/consumer/src/lib.rs
@@ -26,19 +26,19 @@ mod tests {
 
     use crate::FilterResult;
 
-    pub const ROOT_ID: NodeId = NodeId(1);
-    pub const PARAGRAPH_0_ID: NodeId = NodeId(2);
-    pub const STATIC_TEXT_0_0_IGNORED_ID: NodeId = NodeId(3);
-    pub const PARAGRAPH_1_IGNORED_ID: NodeId = NodeId(4);
-    pub const STATIC_TEXT_1_0_ID: NodeId = NodeId(5);
-    pub const PARAGRAPH_2_ID: NodeId = NodeId(6);
-    pub const STATIC_TEXT_2_0_ID: NodeId = NodeId(7);
-    pub const PARAGRAPH_3_IGNORED_ID: NodeId = NodeId(8);
-    pub const EMPTY_CONTAINER_3_0_IGNORED_ID: NodeId = NodeId(9);
-    pub const LINK_3_1_IGNORED_ID: NodeId = NodeId(10);
-    pub const STATIC_TEXT_3_1_0_ID: NodeId = NodeId(11);
-    pub const BUTTON_3_2_ID: NodeId = NodeId(12);
-    pub const EMPTY_CONTAINER_3_3_IGNORED_ID: NodeId = NodeId(13);
+    pub const ROOT_ID: NodeId = NodeId(0);
+    pub const PARAGRAPH_0_ID: NodeId = NodeId(1);
+    pub const STATIC_TEXT_0_0_IGNORED_ID: NodeId = NodeId(2);
+    pub const PARAGRAPH_1_IGNORED_ID: NodeId = NodeId(3);
+    pub const STATIC_TEXT_1_0_ID: NodeId = NodeId(4);
+    pub const PARAGRAPH_2_ID: NodeId = NodeId(5);
+    pub const STATIC_TEXT_2_0_ID: NodeId = NodeId(6);
+    pub const PARAGRAPH_3_IGNORED_ID: NodeId = NodeId(7);
+    pub const EMPTY_CONTAINER_3_0_IGNORED_ID: NodeId = NodeId(8);
+    pub const LINK_3_1_IGNORED_ID: NodeId = NodeId(9);
+    pub const STATIC_TEXT_3_1_0_ID: NodeId = NodeId(10);
+    pub const BUTTON_3_2_ID: NodeId = NodeId(11);
+    pub const EMPTY_CONTAINER_3_3_IGNORED_ID: NodeId = NodeId(12);
 
     pub fn test_tree() -> crate::tree::Tree {
         let mut classes = NodeClassSet::new();
--- a/consumer/src/lib.rs
+++ b/consumer/src/lib.rs
@@ -23,25 +23,22 @@ mod tests {
     use accesskit::{
         Affine, NodeBuilder, NodeClassSet, NodeId, Rect, Role, Tree, TreeUpdate, Vec2,
     };
-    use std::num::NonZeroU128;
 
     use crate::FilterResult;
 
-    pub const ROOT_ID: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(1) });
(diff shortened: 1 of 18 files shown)


consumer/src/lib.rs
```diff
@@ -23,25 +23,22 @@ mod tests {
...
-    pub const STATIC_TEXT_2_0_ID: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(7) });
-    pub const PARAGRAPH_3_IGNORED_ID: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(8) });
-    pub const EMPTY_CONTAINER_3_0_IGNORED_ID: NodeId =
-        NodeId(unsafe { NonZeroU128::new_unchecked(9) });
-    pub const LINK_3_1_IGNORED_ID: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(10) });
-    pub const STATIC_TEXT_3_1_0_ID: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(11) });
-    pub const BUTTON_3_2_ID: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(12) });
-    pub const EMPTY_CONTAINER_3_3_IGNORED_ID: NodeId =
-        NodeId(unsafe { NonZeroU128::new_unchecked(13) });
+    pub const ROOT_ID: NodeId = NodeId(1);
```
DataTriny, line 29: To be consistent with the examples, I guess we want these IDs to start at 0 as well.

consumer/src/node.rs
```diff
@@ -745,16 +745,15 @@ impl Deref for DetachedNode {
...
 
     use crate::tests::*;
 
-    const NODE_ID_1: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(1) });
-    const NODE_ID_2: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(2) });
-    const NODE_ID_3: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(3) });
-    const NODE_ID_4: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(4) });
-    const NODE_ID_5: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(5) });
-    const NODE_ID_6: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(6) });
+    const NODE_ID_1: NodeId = NodeId(1);
```
DataTriny, line 751: To be consistent with the examples, I guess we want these IDs to start at 0 as well.

consumer/src/text.rs
```diff
@@ -1026,16 +1026,15 @@ impl<'a> Node<'a> {
...
-    const NODE_ID_1: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(1) });
-    const NODE_ID_2: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(2) });
-    const NODE_ID_3: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(3) });
-    const NODE_ID_4: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(4) });
-    const NODE_ID_5: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(5) });
-    const NODE_ID_6: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(6) });
-    const NODE_ID_7: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(7) });
-    const NODE_ID_8: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(8) });
+
+    const NODE_ID_1: NodeId = NodeId(1);
```
DataTriny, line 1030: To be consistent with the examples, I guess we want these IDs to start at 0 as well.

consumer/src/tree.rs
```diff
@@ -338,11 +338,10 @@ impl Tree {
 #[cfg(test)]
 mod tests {
     use accesskit::{NodeBuilder, NodeClassSet, NodeId, Role, Tree, TreeUpdate};
-    use std::num::NonZeroU128;
 
-    const NODE_ID_1: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(1) });
-    const NODE_ID_2: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(2) });
-    const NODE_ID_3: NodeId = NodeId(unsafe { NonZeroU128::new_unchecked(3) });
+    const NODE_ID_1: NodeId = NodeId(1);
```
DataTriny, line 342: To be consistent with the examples, I guess we want these IDs to start at 0 as well.

platforms/windows/src/tests/simple.rs
```diff
@@ -3,7 +3,7 @@
 // the LICENSE-APACHE file) or the MIT license (found in
 // the LICENSE-MIT file), at your option.
 
-use std::{convert::TryInto, num::NonZeroU128};
+use std::convert::TryInto;
```
DataTriny, line 6: `TryInto` is in the prelude since Rust 2021 so this can be removed.

DataTriny (changes requested):

DataTriny (approved):

Issue #255, opened 2023-06-24
When should the window be created hidden?


marc2332:
Hey!

While I was working on accessibility support for [Freya](https://github.com/marc2332/freya) I noticed a weird behavior. The initial redraw request event never happens. So after some digging, I found out it only occurs when creating the Winit adapter

I have only tested this on Windows 11.

When creating the Winit adapter:
![image](https://github.com/AccessKit/accesskit/assets/38158676/6e837599-dfab-46b8-bd71-d20e215462a6)

When not creating it :
![image](https://github.com/AccessKit/accesskit/assets/38158676/7f80b368-4b15-4c04-b40a-67aa4c49f952)

I made a little [reproduction](https://github.com/marc2332/bug-accesskit-winit-initial-redraw-request) example, the important line is [this one](https://github.com/marc2332/bug-accesskit-winit-initial-redraw-request/blob/876a069806dd87776476ac6c593533977e1422d6/src/main.rs#L161)

This issue is easy to fix by simply listening to the Init event of the eventloop and forcing a redraw request:

```rust
 Event::NewEvents(StartCause::Init) => {
    window.request_redraw();
}
```

But, is still not the expected behavior and I wonder if it could be causing other problems I am not aware of.

Anyway, nice job with accesskit 👏

mwcampbell:
Thanks for the report. Can you please specify which platform(s) this is happening on?

marc2332:
> Thanks for the report. Can you please specify which platform(s) this is happening on?

I have only tested this in Windows 11, I can try in Linux if you want, but not MacOS as I don't have a Mac 😄

mwcampbell:
You need to create the window invisible, then create the adapter, then show the window. I'm sorry I didn't make that clear in the accesskit_winit docs. This is crucial on Windows, and possibly important on macOS as well.

marc2332:
> You need to create the window invisible, then create the adapter, then show the window. I'm sorry I didn't make that clear in the accesskit_winit docs. This is crucial on Windows, and possibly important on macOS as well.

Ohh, I will try that, Linux works btw :smile:

DataTriny:
Maybe we should panic if `Window::is_visible() == Some(true)` when initializing the adapter.

marc2332:
> You need to create the window invisible, then create the adapter, then show the window. I'm sorry I didn't make that clear in the accesskit_winit docs. This is crucial on Windows, and possibly important on macOS as well.

[...]

PR #278, opened 2023-08-24
refactor!: Decouple in-tree focus from host window/view focus

mwcampbell:
I'm making this change for two reasons. First, I've found that toolkit developers are confused by the existing rules around `TreeUpdate::focus`, and they tend to get it wrong. Second, the old design doesn't fully align with [my proposed vision for a new accessibility architecture for GNOME and other free desktops](https://github.com/mwcampbell/ng-a11y), because it mixes some global state into the tree update.

I've prioritized doing this change now because I already did the `NodeId` breaking change, so I think we should do all anticipated breaking changes at once.

### DIFF
--- a/platforms/winit/src/lib.rs
+++ b/platforms/winit/src/lib.rs
@@ -232,6 +232,9 @@ impl Adapter {
                     Rect::from_origin_size(inner_position, inner_size),
                 )
             }
+            WindowEvent::Focused(is_focused) => {
+                self.adapter.update_window_focus_state(*is_focused);
+            }
             _ => (),
         }
         true
--- a/platforms/winit/src/lib.rs
+++ b/platforms/winit/src/lib.rs
@@ -171,73 +171,9 @@ impl Adapter {
         Self { adapter }
     }
 
-    #[cfg(not(all(
-        feature = "accesskit_unix",
-        any(
-            target_os = "linux",
-            target_os = "dragonfly",
-            target_os = "freebsd",
-            target_os = "netbsd",
-            target_os = "openbsd"
-        )
-    )))]
-    #[must_use]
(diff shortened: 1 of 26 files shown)


platforms/winit/src/lib.rs
```diff
@@ -171,70 +180,9 @@ impl Adapter {
...
-        any(
-            target_os = "linux",
-            target_os = "dragonfly",
-            target_os = "freebsd",
-            target_os = "netbsd",
-            target_os = "openbsd"
-        )
-    ))]
     #[must_use]
     pub fn on_event(&self, window: &Window, event: &WindowEvent) -> bool {
```
DataTriny, line 184: By now I think it is clear that we don't want to prevent the user from processing any winit event, so I think we should take this refactoring as an opportunity to remove `bool` from the return type of this function.

DataTriny (changes requested):

DataTriny:
This PR will fix #255 right?

mwcampbell:
@DataTriny Yes, because the requirement is now explicitly documented for the Windows and macOS subclassing adapters, as well as the winit adapter.

platforms/winit/src/lib.rs
mwcampbell: I'm not sure. We definitely rejected the old way of handling keyboard input with AT-SPI, where we'd have to not only prevent the immediate handling of the event, but asynchronously re-inject the event later. Still, what if we later introduce a self-voicing adapter, basically an embedded screen reader, for environments that don't have a screen reader or accessibility API? Some of those environments might still use a port of winit. In that case, we'd want to prevent the application from handling some keypresses. What do you think?

DataTriny: I wasn't planning this much into the future, but it could be useful indeed.

DataTriny (approved):

PR #281, opened 2023-08-27
Fix doc build for accesskit_winit

dhardy:
Should fix doc build for https://docs.rs/crate/accesskit_winit/

### DIFF
--- a/platforms/winit/Cargo.toml
+++ b/platforms/winit/Cargo.toml
@@ -10,6 +10,9 @@ repository = "https://github.com/AccessKit/accesskit"
 readme = "README.md"
 edition = "2021"
 
+[package.metadata.docs.rs]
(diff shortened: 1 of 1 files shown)


mwcampbell:
@dhardy Please back out the change to accesskit_c. I don't want to trigger a new release of that crate with the changelog entry from this PR.

dhardy:
@mwcampbell the reason for that is a reported conflict of name when building `cargo doc --all` (or similar). But that probably isn't necessary for this exact issue, so will do.

DataTriny:
Yeah I noticed it too... It won't be an issue with docs.rs since this crate is never published, but we will have to fix it.

mwcampbell:
@dhardy Please do open a separate PR with the accesskit_c fix. Thanks.

PR #286, opened 2023-08-28
bindings/c: disable doc

dhardy:
Partial fix for `cargo doc --all --no-deps`.

But on Linux you also need `--features winit/wayland` and then still get an error due to `platforms/macos`.

### DIFF
--- a/bindings/c/Cargo.toml
+++ b/bindings/c/Cargo.toml
@@ -9,6 +9,7 @@ edition = "2021"
 [lib]
 name = "accesskit"
 crate-type = ["cdylib", "staticlib"]
+doc = false
 
 [features]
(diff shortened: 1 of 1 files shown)


DataTriny:
I'm not even sure this is the right way to fix the naming conflict. I initially went with accesskit as the C bindings lib name because accesskit_c.dll is kinda ugly, but I wonder whether we should rename the libs after they are built.

mwcampbell:
I have no problem with disabling Rust docs for this crate. It seems to me that building Rust docs for the C bindings isn't at all useful.

DataTriny (approved):
This PR goes in the right direction of course. Thank you @dhardy

Issue #241, opened 2023-04-19
AT-SPI Accessible is missing get_role_name and get_localized_role_name


wash2:
`get_localized_role_name` is called by accerciser for every node and if it is not implemented, the application and its accessibility tree don't show up at all in accerciser. I've added a simple implementation for each here https://github.com/wash2/accesskit/tree/iced and I'd be happy to make a PR with the changes, but also it might be good to come up with a solution which can localize role names :sweat_smile:

wash2:
Ok gotcha!

DataTriny:
So I made more tests today and my initial assumption was wrong. Given how `GetLocalizedRoleName`'s documentation was phrased, I thought we could return an empty string to signal that libatspi should use its own translation string instead.

Unfortunately it is not the case currently. I'll see if I can change libatspi's behavior, but don't expect a quick fix.

I have a branch in which I updated to the latest atspi version and added the missing methods, `get_localized_role_name` simply returning the role name as provided by the atspi crate. With this our example program shows up in accerciser.

Is it worth adding this broken implementation while we wait for a proper fix? @mwcampbell

DataTriny:
Progress on the GNOME side can be tracked here: [GNOME/at-spi2-core#110](https://gitlab.gnome.org/GNOME/at-spi2-core/-/issues/110).

DataTriny:
Underlying AT-SPI issue was closed. Our winit example now shows up on Accerciser when using at-spi2-core master. Once you have cloned [the repo](https://gitlab.gnome.org/GNOME/at-spi2-core), instructions for how to build and install can be found [here](https://gitlab.gnome.org/GNOME/at-spi2-core/-/blob/main/INSTALL).

Now of course this mean that it won't work for anyone using the current version. Given the state of AT-SPI and how much it is predicted to change, doing backports on our side will be a nightmare. If compatibility with older GNOME is a feature we want, then this issue is still relevant, otherwise it can be closed.

lunixbochs:
My vote would be at minimum for compatibility with any currently supported ubuntu LTS, so I'd prefer a workaround until that atspi version is EOL.

lunixbochs:
I wouldn't mind having this mitigated, as it currently means anyone with an existing linux install can't use Accerciser from their package manager to test Accesskit apps.

DataTriny:
[...]

PR #290, opened 2023-08-31
don't require tokio rt-multi-thread feature

lunixbochs:


DataTriny:
Hello @lunixbochs 
Please run `cargo fmt` on your branch.

lunixbochs:
done

mwcampbell:
How exactly does this work? Why is it necessary to spawn a new thread, create a tokio runtime on that thread, send the runtime back, and then exit the thread that created it?

lunixbochs:
Wow, I think I was completely on autopilot. I'd only done basic compile+println testing because existing versions of Accerciser still don't work with Accesskit on Linux due to #241

Sending a handle back and parking the thread on a sleep works much better. I tested this approach further by manually connecting to the bus and verified some atspi methods work.

mwcampbell:
This approach looks good to me. I just belatedly approved the CI workflow.

DataTriny:
The winit example seem to still work properly with the `tokio` feature on.

lunixbochs:
I just realized I can send the runtime into the thread on creation, which simplifies the code quite a bit.

PR #291, opened 2023-09-01
feat!: Allow providing app_name, toolkit_name and toolkit_version in Tree, remove parameters from unix adapter constructor

### LINKED ISSUE #280: Implement UIA provider properties
For cleanliness, I think it would be nice if UIA properties would be implemented that allow UIA elements to be identified as coming from accesskit:

* FrameworkIdProperty: Probably just `AccessKit`
* ProviderDescriptionProperty: I'm not sure how this one is used, but it might be beneficial to provide version information of the accesskit_windows crate

Probably pretty trivial, but I think this can could be considered an issue a new developer can start with 😉
I guess there can some debate about whether an user of the crate can update both properties, for example to communicate that the element is coming from their framework.

leonardder:
Similar to the unix crate, this adds `app_name`, `toolkit_name` and `toolkit_version` parameters to the adapter constructors.
- FrameworkIdProperty: exposes `toolkit_name`
- ProviderDescriptionProperty: exposes `app_name (toolkit_name toolkit_version)`

fixes #280

### DIFF
--- a/bindings/c/src/common.rs
+++ b/bindings/c/src/common.rs
@@ -864,8 +864,8 @@ impl BoxCastPtr for tree {}
 impl tree {
     #[no_mangle]
     pub extern "C" fn accesskit_tree_new(root: node_id) -> *mut tree {
-        let t = Tree::new(root.into());
-        BoxCastPtr::to_mut_ptr(t)
+        let internal_tree = Tree::new(root.into());
+        BoxCastPtr::to_mut_ptr(internal_tree)
     }
 
     #[no_mangle]
@@ -882,15 +882,19 @@ impl tree {
     }
 
     #[no_mangle]
-    pub extern "C" fn accesskit_tree_set_toolkit_info(
-        t: *mut tree,
-        toolkit_name: *const c_char,
-        toolkit_version: *const c_char,
-    ) {
+    pub extern "C" fn accesskit_tree_set_toolkit_name(t: *mut tree, toolkit_name: *const c_char) {
         let t = mut_from_ptr(t);
         t.toolkit_name = Some(String::from(
             unsafe { CStr::from_ptr(toolkit_name) }.to_string_lossy(),
         ));
+    }
+
+    #[no_mangle]
+    pub extern "C" fn accesskit_tree_set_toolkit_version(
+        t: *mut tree,
+        toolkit_version: *const c_char,
+    ) {
+        let t = mut_from_ptr(t);
         t.toolkit_version = Some(String::from(
             unsafe { CStr::from_ptr(toolkit_version) }.to_string_lossy(),
         ));
--- a/bindings/c/src/common.rs
+++ b/bindings/c/src/common.rs
@@ -875,10 +875,10 @@ impl tree {
 
     /// Caller must call `accesskit_string_free` with the return value.
     #[no_mangle]
-    pub extern "C" fn accesskit_tree_get_app_name(tree: *mut tree) -> *mut c_char {
-        let tree = box_from_ptr(tree);
-        match tree.app_name {
-            Some(value) => CString::new(value).unwrap().into_raw(),
+    pub extern "C" fn accesskit_tree_get_app_name(tree: *const tree) -> *mut c_char {
+        let tree = ref_from_ptr(tree);
+        match tree.app_name.as_ref() {
+            Some(value) => CString::new(value.clone()).unwrap().into_raw(),
             None => ptr::null_mut(),
         }
     }
@@ -899,10 +899,10 @@ impl tree {
 
     /// Caller must call `accesskit_string_free` with the return value.
     #[no_mangle]
-    pub extern "C" fn accesskit_tree_get_toolkit_name(tree: *mut tree) -> *mut c_char {
-        let tree = box_from_ptr(tree);
-        match tree.toolkit_name {
-            Some(value) => CString::new(value).unwrap().into_raw(),
+    pub extern "C" fn accesskit_tree_get_toolkit_name(tree: *const tree) -> *mut c_char {
+        let tree = ref_from_ptr(tree);
+        match tree.toolkit_name.as_ref() {
+            Some(value) => CString::new(value.clone()).unwrap().into_raw(),
             None => ptr::null_mut(),
         }
     }
@@ -926,10 +926,10 @@ impl tree {
 
     /// Caller must call `accesskit_string_free` with the return value.
     #[no_mangle]
-    pub extern "C" fn accesskit_tree_get_toolkit_version(tree: *mut tree) -> *mut c_char {
-        let tree = box_from_ptr(tree);
-        match tree.toolkit_version {
-            Some(value) => CString::new(value).unwrap().into_raw(),
+    pub extern "C" fn accesskit_tree_get_toolkit_version(tree: *const tree) -> *mut c_char {
+        let tree = ref_from_ptr(tree);
+        match tree.toolkit_version.as_ref() {
+            Some(value) => CString::new(value.clone()).unwrap().into_raw(),
             None => ptr::null_mut(),
         }
     }
--- a/bindings/c/src/common.rs
+++ b/bindings/c/src/common.rs
@@ -851,23 +851,53 @@ impl node_builder {
     }
 }
 
-#[repr(C)]
 pub struct tree {
-    pub root: node_id,
+    _private: [u8; 0],
 }
+
+impl CastPtr for tree {
+    type RustType = Tree;
+}
+
+impl BoxCastPtr for tree {}
 
 impl tree {
     #[no_mangle]
-    pub extern "C" fn accesskit_tree_new(root: node_id) -> tree {
-        tree { root }
+    pub extern "C" fn accesskit_tree_new(root: node_id) -> *mut tree {
+        let t = Tree::new(root.into());
+        BoxCastPtr::to_mut_ptr(t)
     }
-}
 
-impl From<tree> for Tree {
-    fn from(tree: tree) -> Self {
-        Self {
-            root: tree.root.into(),
-        }
+    #[no_mangle]
+    pub extern "C" fn accesskit_tree_free(t: *mut tree) {
+        drop(box_from_ptr(t));
+    }
+
+    #[no_mangle]
+    pub extern "C" fn accesskit_tree_set_app_name(t: *mut tree, app_name: *const c_char) {
+        let t = mut_from_ptr(t);
+        t.app_name = Some(String::from(
+            unsafe { CStr::from_ptr(app_name) }.to_string_lossy(),
+        ));
+    }
+
+    #[no_mangle]
+    pub extern "C" fn accesskit_tree_set_toolkit_name(t: *mut tree, toolkit_name: *const c_char) {
+        let t = mut_from_ptr(t);
+        t.toolkit_name = Some(String::from(
+            unsafe { CStr::from_ptr(toolkit_name) }.to_string_lossy(),
+        ));
+    }
+
+    #[no_mangle]
+    pub extern "C" fn accesskit_tree_set_toolkit_version(
+        t: *mut tree,
+        toolkit_version: *const c_char,
+    ) {
+        let t = mut_from_ptr(t);
+        t.toolkit_version = Some(String::from(
+            unsafe { CStr::from_ptr(toolkit_version) }.to_string_lossy(),
+        ));
     }
 }
 
@@ -924,9 +954,9 @@ impl tree_update {
     }
 
     #[no_mangle]
-    pub extern "C" fn accesskit_tree_update_set_tree(update: *mut tree_update, tree: tree) {
+    pub extern "C" fn accesskit_tree_update_set_tree(update: *mut tree_update, tree: *mut tree) {
         let update = mut_from_ptr(update);
-        update.tree = Some(tree.into());
+        update.tree = Some(*box_from_ptr(tree));
     }
 
     #[no_mangle]
--- a/bindings/c/src/common.rs
+++ b/bindings/c/src/common.rs
@@ -882,19 +882,15 @@ impl tree {
     }
 
     #[no_mangle]
-    pub extern "C" fn accesskit_tree_set_toolkit_name(t: *mut tree, toolkit_name: *const c_char) {
+    pub extern "C" fn accesskit_tree_set_toolkit_info(
+        t: *mut tree,
+        toolkit_name: *const c_char,
+        toolkit_version: *const c_char,
+    ) {
         let t = mut_from_ptr(t);
         t.toolkit_name = Some(String::from(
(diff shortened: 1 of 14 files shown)


DataTriny:
Thanks @leonardder for taking this!

I think the adapters constructors have too many arguments now. I didn't put too much thought into the Unix adapter because I wasn't sure what we would do with this information, but if we support it for Windows too, I think we should consider the builder pattern, and preferrably do it in a similar way on both Windows and Unix.

I don't know if we can implement the same for macOS, but now might be the right time to allow setting these values for the winit adapter as well, the builder pattern might be relevant here too.

DataTriny:
Also before you push here, please make sure to format your code by running `cargo fmt` (on the whole workspace) and `cargo clippy` (on the updated crates).

mwcampbell:
I was actually thinking about putting this info in the `accesskit::Tree` struct that has to be provided with the first tree update. That might complicate how that structure is built in the C bindings, but I think it would make sense.

DataTriny:
This would indeed make sense. Do we want these fields to be optional?

mwcampbell:
Yes, definitely optional. So make them `Option<String>`, and if they're absent, then the Windows adapter leaves those properties empty (as in an empty variant, not an empty string), and the Unix adapter returns empty strings.

leonardder:
> I was actually thinking about putting this info in the `accesskit::Tree` struct that has to be provided with the first tree update. That might complicate how that structure is built in the C bindings, but I think it would make sense.

I'm happy to give this a try, but I assume that also means removing the constructor parameters from the unix adapter right? If so, it might be better to close this pr and come up with something new whenever I'm done with that.

mwcampbell:
@leonardder We definitely want to drop those options from the Unix adapter's constructor. For that reason, I'd like to include this change alongside the other breaking changes we've been making lately, if you have time to do it in the next week or so. Whether you close this PR and open a new one, or just edit this one, is up to you. It's only a draft right now anyway.

mwcampbell:
As for the C bindings, I think we'll need to turn `accesskit_tree` into an opaque boxed type with setters, like I recently did for `accesskit_tree_update`.

leonardder:
> For that reason, I'd like to include this change alongside the other breaking changes we've been making lately, if you have time to do it in the next week or so.

I'll do whatever I can.

leonardder:
I added three new fields to the Tree struct. We can either:
1. Expose them in the `new` constructor method
2. Remove the `new` constructor (i.e. use struct initialization)
3. Leave them optional at construction time. This keeps backwards compatibility but if someone wants to fill the struct field, he ought to store the tree instance in a separate variable first.

mwcampbell:
My preferences is to keep these fields optional on `Tree` and keep the current constructor. I'm willing to be convinced otherwise though.

leonardder:
I think I share that preference, so I'll go ahead wit that. Thanks for the quick response!

leonardder:
I think this is ready to rock now.

DataTriny:
Yeah I encountered the same issue with consts, their support in initializers is quite recent apparently. We should maybe switch to using `#define` as to avoid initializing the initial focus at runtime.

bindings/c/src/common.rs
```diff
@@ -851,23 +851,49 @@ impl node_builder {
...
+    #[no_mangle]
+    pub extern "C" fn accesskit_tree_set_app_name(t: *mut tree, app_name: *const c_char) {
+        let t = mut_from_ptr(t);
+        t.app_name = Some(String::from(
+            unsafe { CStr::from_ptr(app_name) }.to_string_lossy(),
+        ));
+    }
+
+    #[no_mangle]
+    pub extern "C" fn accesskit_tree_set_toolkit_info(
```
DataTriny, line 885: I understand why you chose to have one function here, I'm in favor of keeping the C bindings as close to the Rust schema as possible though.

bindings/c/src/common.rs
```diff
@@ -851,23 +851,49 @@ impl node_builder {
...
-        Self {
-            root: tree.root.into(),
-        }
+    #[no_mangle]
+    pub extern "C" fn accesskit_tree_free(t: *mut tree) {
+        drop(box_from_ptr(t));
+    }
+
+    #[no_mangle]
+    pub extern "C" fn accesskit_tree_set_app_name(t: *mut tree, app_name: *const c_char) {
```
DataTriny, line 877: For consistency with other types, having get and clear functions for each property would be nice. I'm also not a big fan of one-letter variables, especially arguments.

bindings/c/src/common.rs
```diff
@@ -924,9 +950,9 @@ impl tree_update {
     }
 
     #[no_mangle]
-    pub extern "C" fn accesskit_tree_update_set_tree(update: *mut tree_update, tree: tree) {
+    pub extern "C" fn accesskit_tree_update_set_tree(update: *mut tree_update, tree: *mut tree) {
```
DataTriny, line 953: Would be good to have at least `clear_tree` here as well.

common/src/lib.rs
```diff
@@ -2236,13 +2236,28 @@ impl JsonSchema for Node {
 #[cfg_attr(feature = "serde", serde(deny_unknown_fields))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 pub struct Tree {
+    /// The identifier of the tree's root node
     pub root: NodeId,
+    /// The name of the application this tree belongs to.
+    /// This value is optional.
```
DataTriny, line 2242: No need to say that it's optional, the type system already communicates this fact.

common/src/lib.rs
```diff
@@ -2236,13 +2236,28 @@ impl JsonSchema for Node {
...
+    pub app_name: Option<String>,
```
DataTriny, line 2243: Thinking more about this, I wonder if we shouldn't require the app name afterall. Some UI toolkits/game engines force users to provide one to manage config files and stuff, and without the app name ATs wouldn't have a reliable way to patch an app.

Just a thought, feel free to disagree.

common/src/lib.rs
```diff
@@ -2236,13 +2236,28 @@ impl JsonSchema for Node {
...
+    /// The name of the used UI toolkit.
```
DataTriny, line 2244:
```suggestion
    /// Name of the UI toolkit in use.
```

platforms/unix/src/adapter.rs
```diff
@@ -157,8 +154,14 @@ impl Adapter {
             },
             "accesskit_event_task",
         );
-        let tree = Tree::new(initial_state(), is_window_focused);
-        let app_context = AppContext::new(app_name, toolkit_name, toolkit_version);
+        let initial_state = initial_state();
+        let initial_tree = initial_state.tree.as_ref().unwrap();
+        let app_context = AppContext::new(
+            initial_tree.app_name.clone().unwrap_or_default(),
```
DataTriny, line 160: I don't really know why would someone do this, but we allow changing these values at any time. To be accurate, we should check on each tree update if the values are provided, and update them accordingly.

DataTriny (changes requested):
I haven't reviewed everything yet.

I'll test the examples on Linux tomorrow if noone can do it before then.

common/src/lib.rs
leonardder: I'm unopinionated about this. I'm also not sure how we'd use the app_name on macOS yet.

platforms/unix/src/adapter.rs
leonardder: I have considered removing the fields from the app_context, since they are now on the tree anyway. What do you think about that? We can just always get them from the current Tree instance and then we wouldn't have to care about updating the context. just a thought though.

DataTriny: This is actually colliding with #292.
What if two adapters push different values? @mwcampbell

mwcampbell: I think the best we can do in this case is what #292 already does; the first set of values wins.

mwcampbell: Sorry, I had overlooked the context of the comment. For the Unix adapter, I think we want to keep the fields on the app context, and copy them there from the first tree update when the app context is created.

consumer/src/tree.rs
```diff
@@ -258,6 +258,45 @@ impl State {
...
+
+    pub fn toolkit_name(&self) -> Option<String> {
+        self.data.toolkit_name.clone()
+    }
+
+    pub fn toolkit_version(&self) -> Option<String> {
+        self.data.toolkit_version.clone()
+    }
+
+    pub fn app_and_toolkit_description(&self) -> Option<String> {
```
DataTriny, line 274: This function seem too Windows specific to stay here at the moment. If another adapter end up needing it then we can put it inside the consumer crate but I'd suggest moving it into `accesskit_windows` for now.

consumer/src/tree.rs
```diff
@@ -258,6 +258,45 @@ impl State {
...
+        let app_name = self.app_name();
+        let toolkit_name = self.toolkit_name();
+        let toolkit_version = self.toolkit_version();
+        if let (Some(app_name), Some(toolkit_name), Some(toolkit_version)) =
```
DataTriny, line 278: I think that a `match` statement would be more readable.

common/src/lib.rs
```diff
@@ -2236,13 +2236,28 @@ impl JsonSchema for Node {
 #[cfg_attr(feature = "serde", serde(deny_unknown_fields))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 pub struct Tree {
+    /// The identifier of the tree's root node
     pub root: NodeId,
+    /// The name of the application this tree belongs to.
+    /// This value is optional.
+    pub app_name: Option<String>,
+    /// Name of the UI toolkit in use.
```
DataTriny, line 2244:
```suggestion
    /// The name of the UI toolkit in use.
```

bindings/c/src/common.rs
DataTriny: I would just go with `tree`.

DataTriny (changes requested):

bindings/c/src/common.rs
leonardder: A, I thought having a struct and variable with the same name would fail, but turns out it doesn't.

I'm not very up to speed with macros yet. DO you suggest using macros or should I just create the getters and clearers manually?

DataTriny: The macros we currently have can't be reused, and it would not be worth to define new ones just for a struct. Hand-writting these functions seem the best option to me.

platforms/unix/src/adapter.rs
leonardder: Copying the values at initialization time is exactly what happens now, right? I'm not sure whether I should still change anything here.

platforms/windows/examples/hello_world.rs
```diff
@@ -105,13 +105,18 @@ impl InnerWindowState {
         let root = self.build_root();
         let button_1 = build_button(BUTTON_1_ID, "Button 1", &mut self.node_classes);
         let button_2 = build_button(BUTTON_2_ID, "Button 2", &mut self.node_classes);
+        let mut tree = Tree::new(WINDOW_ID);
+        tree.app_name = Some(WINDOW_TITLE.to_string());
+        tree.toolkit_name = Some(env!("CARGO_PKG_NAME").to_string());
```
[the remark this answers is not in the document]
leonardder: I will remove them from the examples for now.

leonardder:
I believe I have addressed all the review comments now.
I changed the offending const in the examples to a #define but didn't touch the other constants.

platforms/unix/src/adapter.rs
DataTriny: No you're good.

bindings/c/src/common.rs
```diff
@@ -924,9 +1005,9 @@ impl tree_update {
...
```
DataTriny, line 1008: I think we still want a `clear_tree` function.

platforms/winit/examples/simple.rs
```diff
@@ -83,13 +83,17 @@ impl State {
         let root = self.build_root();
         let button_1 = build_button(BUTTON_1_ID, "Button 1", &mut self.node_classes);
         let button_2 = build_button(BUTTON_2_ID, "Button 2", &mut self.node_classes);
+        let mut tree = Tree::new(WINDOW_ID);
+        tree.app_name = Some(env!("CARGO_BIN_NAME").to_string());
+        tree.toolkit_name = Some(String::from("winit"));
+        tree.toolkit_version = Some(String::from("0.28"));
```
DataTriny, line 89: We'll now have to manually update this everytime we bump winit's version. winit isn't even the UI toolkit here, just a windowing library. I'm now pretty sure it's better to remove toolkit name and version from this example and the SDL one.

DataTriny (changes requested):
Great work overall!
Probably the last comments for me.

bindings/c/src/common.rs
DataTriny: Indeed it was... Sorry about that.

common/src/lib.rs
```diff
@@ -2236,13 +2236,25 @@ impl JsonSchema for Node {
 #[cfg_attr(feature = "serde", serde(deny_unknown_fields))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 pub struct Tree {
+    /// The identifier of the tree's root node
```
DataTriny, line 2239:
```suggestion
    /// The identifier of the tree's root node.
```

DataTriny (changes requested):

DataTriny:
I've applied a non-controversial change and I rebased this to save you the hassle. For me this PR is OK now.

Thank you very much for working on this.

leonardder:
You're very welcome! Thanks for the rebase!

mwcampbell:
@leonardder You should definitely add your name to the AUTHORS file.

I think it would be a good idea to rebase this branch onto the latest main commit one more time, just to make sure none of the recent Unix adapter changes introduce conflicts.

Please make sure you've reformatted the C examples with clang-format.

mwcampbell:
Otherwise, this looks good. And, having read the minor changes to the Unix adapter, I doubt that we need another rebase now.

leonardder:
> Please make sure you've reformatted the C examples with clang-format.

I have performed formatting with VS Code, I'm pretty sure that uses a clangformat file when available, but I'll double check.

leonardder:
I have merged in main and made sure that the c files are formatted correctly. I assume you'll perform a squash merge?

mwcampbell:
@leonardder Yes, I always do a squash merge.

Now I just need you to add your name to the AUTHORS file, and then once the CI passes, I'm ready to merge this. Thanks for your contribution!

leonardder:
> Now I just need you to add your name to the AUTHORS file, and then once the CI passes, I'm ready to merge this. 

AH I'm sorry, Done now.

> Thanks for your contribution!

You're very welcome!

PR #292, opened 2023-09-03
feat: Support multiple top-level windows on Unix

DataTriny:
This PR includes the following changes:

- Adapters now use a single D-Bus connection: they all update the same D-Bus tree and the app has one D-Bus unique name,
- Adapters are now identified by an ID, which gets passed to platform nodes,
- `AppContext` keep track of all the adapters: the `Application` interface can now report the list of top-level windows,
- Signals are emitted when windows are added, removed and become active.
- The `Active` state of top-level windows is now only set if the window is actually the active one.


mwcampbell:
I only see one problem with this implementation. I think the adapter's `Drop` implementation should remove the adapter from the app context's list. Otherwise, this is a good solution.

DataTriny:
@mwcampbell I think this is now ready.

mwcampbell:
@DataTriny One of your recent commits apparently caused platforms/unix/src/atspi/interfaces/accessible.rs to be rewritten with CRLF line endings. I don't care which form of line endings we use for any given file; I've probably been inconsistent on that myself. But let's not change the format of an existing file, as that adds needlessly to the diff. Please switch the file back to LF line endings. Thanks.

DataTriny:
Sorry about that. I think it is fixed now.

PR #301, opened 2023-10-25
chore: Fix some clippy warnings

DataTriny:
I don't feel confident in how I fixed this warning, from a lint introduced in 1.72:

```
warning: usage of an `Arc` that is not `Send` or `Sync`
   --> platforms\windows\src\tests\mod.rs:273:9
    |
273 | /         Arc::new(Self {
274 | |             mutex: Mutex::new(None),
275 | |             cv: Condvar::new(),
276 | |         })
    | |__________^
    |
    = note: the trait `Send` is not implemented for `ReceivedFocusEvent`
    = note: the trait `Sync` is not implemented for `ReceivedFocusEvent`
    = note: required for `Arc<ReceivedFocusEvent>` to implement `Send` and `Sync`
    = help: consider using an `Rc` instead or wrapping the inner type with a `Mutex`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync
    = note: `#[warn(clippy::arc_with_non_send_sync)]` on by default
```

I don't want to just ignore it, but I don't think we can send `IUIAutomationElement` across threads so easily.

### DIFF
--- a/platforms/windows/examples/hello_world.rs
+++ b/platforms/windows/examples/hello_world.rs
@@ -31,8 +31,7 @@ static WINDOW_CLASS_ATOM: Lazy<u16> = Lazy::new(|| {
 
     let atom = unsafe { RegisterClassW(&wc) };
     if atom == 0 {
-        let result: Result<()> = Err(Error::from_win32());
-        result.unwrap();
+        panic!("{}", Error::from_win32());
     }
     atom
 });
--- a/platforms/windows/src/tests/mod.rs
+++ b/platforms/windows/src/tests/mod.rs
@@ -8,7 +8,7 @@ use once_cell::{sync::Lazy as SyncLazy, unsync::Lazy};
 use std::{
     cell::RefCell,
     rc::Rc,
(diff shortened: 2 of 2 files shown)


mwcampbell (approved):

kpreid:
`unsafe impl Send ...` is _not_ an appropriate fix for the _warning_ unless specifically confirmed to be correct in context. An unnecessary `Arc` may be less efficient than ideal, or misleading, in which case it would ideally be replaced with `Rc`, but that's all the clippy lint is trying to say. An incorrect `unsafe impl Send` is **unsound**.

(I'm just a bystander here, but this is not the first time I've seen someone misunderstand this lint. **Adding unsafe impls should not be done casually.**)

mwcampbell:
@kpreid Good catch. I'm fine with using `Rc` if that doesn't introduce any errors.

DataTriny:
Given the context I didn't even try putting an `Rc`, yet it compiles and the tests seem to pass. Thank you very much @kpreid, I knew I was doing something awful here.

PR #306, opened 2023-11-04
fix: Make the test focus event handler thread-safe again

mwcampbell:
I realized that the recent switch from `Arc` to `Rc` in this event handler was incorrect, because the event handler is in fact called from a different thread from the one on which it was created. To address the original clippy warning in the safest way possible, I created a new wrapper struct, `SendableUiaElement`, that implements `Send`. The compiler can automatically implement `Send` and `Sync` on `ReceivedFocusEvent` from there. I also added a static assertion that `Send` and `Sync` are implemented on `FocusEventHandler`.


DataTriny (approved):
The newtype struct really make things clearer.

PR #307, opened 2023-11-04
fix!: Rename `accesskit_winit::Adapter::on_event` to `process_event`

### LINKED ISSUE #300: access_kit_winit Adapter::on_event is confusingly named and undocumented
[This call](https://github.com/AccessKit/accesskit/blob/a72f90fbdc479c2e0d44074f08464b9ab72ae245/platforms/winit/src/platform_impl/macos.rs#L37) seems to propagate winit events into access_kit in order to be handled correctly.

The boolean returned seems to reflect whether or not the event should be filtered and absorbed, per https://github.com/bevyengine/bevy/pull/10239#issuecomment-1775572176, but currently always returns true across platforms. This is particularly confusing because the wrapping struct (of the same name)'s [method](https://github.com/AccessKit/accesskit/blob/a72f90fbdc479c2e0d44074f08464b9ab72ae245/platforms/winit/src/lib.rs#L78) is marked `#[must_use]`.

I'm new to accesskit's codebase, but a few requests to ease integration:

1. Rename `on_event` to something that clearly indicates that there will be side-effects, like `process_winit_event`
2. Add documentation for these methods.
3. Change the returned `bool` to a more meaningful enum, either a Result or a custom `ShouldFilter` enum to better communicate intended use.

DataTriny:
Closes #300 

The name `on_event` suggests that this function won't have side effects, which is not the case.

We received strong push-back from developers who don't want AccessKit to filter events. We previously agreed on keeping the mechanism around even though we didn't use it for keyboard events on Unix. Even though we might need it in the future for other features, this is not currently the case, therefore it is hard to convince people to use it. I propose to remove it for now.

### DIFF
--- a/platforms/winit/examples/simple.rs
+++ b/platforms/winit/examples/simple.rs
@@ -177,37 +177,40 @@ fn main() -> Result<(), impl std::error::Error> {
         event_loop.set_control_flow(ControlFlow::Wait);
 
         match event {
-            Event::WindowEvent { event, .. } if adapter.on_event(&window, &event) => match event {
-                WindowEvent::CloseRequested => {
-                    event_loop.exit();
-                }
-                WindowEvent::KeyboardInput {
-                    event:
-                        KeyEvent {
-                            logical_key: virtual_code,
-                            state: ElementState::Pressed,
-                            ..
-                        },
-                    ..
-                } => match virtual_code {
-                    Key::Named(winit::keyboard::NamedKey::Tab) => {
-                        let mut state = state.lock().unwrap();
-                        let new_focus = if state.focus == BUTTON_1_ID {
-                            BUTTON_2_ID
-                        } else {
-                            BUTTON_1_ID
-                        };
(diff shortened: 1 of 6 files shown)


DataTriny:
The two same tests are failing again. Looks like #306 didn't change anything afterall.

mwcampbell:
OK, we'll revisit event filtering if and when we actually implement a self-voicing adapter.

alice-i-cecile:
Thanks, y'all rock <3

mwcampbell:
@alice-i-cecile What are Bevy's plans for updating to winit 0.29? I was just wondering if we need to backport this fix to another accesskit_winit release that still uses winit 0.28.

alice-i-cecile:
We'll be upgraded to 0.29 by the next Bevy release. It's a complex migration, but hotly desired. @Vrixyz has been leading those efforts

mwcampbell:
OK, then I gather it's not necessary to backport this API fix for winit 0.28.

PR #308, opened 2023-11-04
fix: Make the Windows subclassing test more robust

mwcampbell:
I see from a [recent failed CI job](https://github.com/AccessKit/accesskit/actions/runs/6755825086/job/18364643052) that the subclassing test can also intermittently fail. I noticed that my own test broke the cardinal rule of using the subclassing adapter: the window must be created invisible. So I fixed that, but I also made a couple of changes:

* The test no longer repeatedly calls `UiaHasServerSideProvider`, and it doesn't drop and re-create the adapter, because these things aren't testing realistic scenarios. In real use, the window is created invisible, the adapter is created once afterward, and nobody would call `UiaHasServerSideProvider` (or equivalent) on the window until after the adapter is created. I've seen that sometimes `UiaHasServerSIdeProvider` appears to cache the result of an earlier call, so it's safer to only call it the one time.
* We actually never show the window, because that's not strictly necessary for this test, and it might actually be what's causing the focus test to intermittently fail in CI, since the focus test can run concurrently with this one. Let's see if this fixes that intermittent failure.

### DIFF
--- a/platforms/windows/src/tests/mod.rs
+++ b/platforms/windows/src/tests/mod.rs
@@ -185,7 +185,7 @@ impl Scope {
 }
 
 // It's not safe to run these UI-related tests concurrently.
-static MUTEX: Mutex<()> = Mutex::new(());
+pub(crate) static MUTEX: Mutex<()> = Mutex::new(());
 
 pub(crate) fn scope<F>(
     window_title: &str,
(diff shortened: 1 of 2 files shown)


mwcampbell:
This time, only the subclassing test failed. I think I have to prevent the subclassing test from running concurrently with the other tests, just as the tests that use the full generic test harness are prevented from running concurrently.

DataTriny (approved):

PR #305, opened 2023-10-29
chore: release main

mwcampbell:
:robot: I have created a release *beep* *boop*
---

<details><summary>accesskit: 0.12.1</summary>

### Bug Fixes

* Add missing semicolons when not returning anything ([#303](https://github.com/AccessKit/accesskit/issues/303)) ([38d4de1](https://github.com/AccessKit/accesskit/commit/38d4de1442247e701047d75122a9638a2ed99b1f))
</details>

<details><summary>accesskit_c: 0.6.1</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * accesskit bumped from 0.12.0 to 0.12.1
    * accesskit_windows bumped from 0.15.0 to 0.15.1
    * accesskit_macos bumped from 0.10.0 to 0.10.1
    * accesskit_unix bumped from 0.6.0 to 0.6.1
</details>

<details><summary>accesskit_consumer: 0.16.1</summary>

### Bug Fixes

* Add missing semicolons when not returning anything ([#303](https://github.com/AccessKit/accesskit/issues/303)) ([38d4de1](https://github.com/AccessKit/accesskit/commit/38d4de1442247e701047d75122a9638a2ed99b1f))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * accesskit bumped from 0.12.0 to 0.12.1
</details>

<details><summary>accesskit_macos: 0.10.1</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * accesskit bumped from 0.12.0 to 0.12.1
    * accesskit_consumer bumped from 0.16.0 to 0.16.1
</details>

<details><summary>accesskit_unix: 0.6.1</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * accesskit bumped from 0.12.0 to 0.12.1
    * accesskit_consumer bumped from 0.16.0 to 0.16.1
</details>

<details><summary>accesskit_windows: 0.15.1</summary>

### Bug Fixes

* Make the test focus event handler thread-safe again ([#306](https://github.com/AccessKit/accesskit/issues/306)) ([bb479c6](https://github.com/AccessKit/accesskit/commit/bb479c69bddff77a1102549a7ff6ea5c7155d90d))
* Make the Windows subclassing test more robust ([#308](https://github.com/AccessKit/accesskit/issues/308)) ([0078f79](https://github.com/AccessKit/accesskit/commit/0078f7997f11a3bb1f3c33dc7a91f427b1a60db8))

[...]

PR #310, opened 2023-11-04
fix: Use raw-window-handle 0.6

mwcampbell:
Following up on an [earlier comment](https://github.com/AccessKit/accesskit/pull/305#issuecomment-1793515487), I don't think it would be right for us to do a new release that depends on an old version of raw-window-handle, if we can easily avoid it. So this PR switches accesskit_winit to use rwh_06. @Vrixyz I believe this can coexist with your code that requires rwh_05, since it seems that those winit features don't actually conflict, but please check.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,7 +94,6 @@ dependencies = [
  "accesskit_macos",
  "accesskit_unix",
  "accesskit_windows",
- "raw-window-handle 0.5.2",
  "winit",
 ]
 
@@ -1123,8 +1122,7 @@ dependencies = [
  "log",
  "ndk-sys",
  "num_enum",
- "raw-window-handle 0.5.2",
- "raw-window-handle 0.6.0",
+ "raw-window-handle",
  "thiserror",
 ]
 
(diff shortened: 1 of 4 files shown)


DataTriny (approved):
The code is OK for me indeed!

Mxing raw-window-handle versions seem to never be an option from the GitHub issues I checked on various projects. Therefore I would prefer more input from users before merging.

Vrixyz:
Yes, I believe your understanding is correct, 2 different versions of rwh should be able to coexist.

For the record, I initially didn't update to rwh_06 because wgpu wasn't supporting it either, so it gave me enough incentive to not update.

by next bevy version (0.13 ; I guess ~march 2024), we'll probably have an updated wgpu version and no blockers to support only rwh_06, I don't think I'll champion those changes, but that should ease the integration if accesskit upgrades to rwh_06.

mwcampbell:
Since winit itself has rwh_06 as a default feature now, I think it's better for us to use that version. Going to merge.

Vrixyz:
Hey, with [bevy winit/accesskit update](https://github.com/bevyengine/bevy/pull/10702), I realized having multiple version is not pleasing our CI (we're trying to avoid duplicate dependencies). Would you be against having a feature `rwh_05` ?

Pretty much reverting your changes behind a feature flag akin to [how winit do it](https://github.com/rust-windowing/winit/pull/3126).

PR #316, opened 2023-12-03
feat: Support custom role descriptions

DataTriny:
Revamping an old branch, initial work by @mwcampbell.

Implement support for the `role_description` property on all adapters:
- Documented the fact that this property should be used sparingly,
- Tweaked the macOS and Windows implementations after rebase,
- On Unix: implemented the `GetLocalizedRoleName` method of the `org.a11y.atspi.Accessible` interface. Make use of `AtspiRole::Extended` if a value is supplied, as it is the only way to tell libatspi to query it. Orca still won't announce it because it currently explicitly ignore this role, Accerciser however displays it properly.

Tested on all platforms by modifying the winit example.

### DIFF
--- a/consumer/src/node.rs
+++ b/consumer/src/node.rs
@@ -358,6 +358,10 @@ impl NodeState {
         self.data().role_description().map(String::from)
     }
 
+    pub fn has_role_description(&self) -> bool {
+        self.data().role_description().is_some()
+    }
+
     pub fn is_hidden(&self) -> bool {
         self.data().is_hidden()
     }
--- a/consumer/src/node.rs
+++ b/consumer/src/node.rs
@@ -354,6 +354,10 @@ impl NodeState {
         self.data().role()
     }
 
+    pub fn role_description(&self) -> Option<String> {
+        self.data().role_description().map(String::from)
+    }
+
     pub fn is_hidden(&self) -> bool {
         self.data().is_hidden()
     }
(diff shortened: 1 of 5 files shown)


mwcampbell:
@DataTriny Are you sure that returning `AtSpiRole::Extended` in the Unix adapter for all nodes with role descriptions is the correct behavior? It would make more sense to me to do that only if the base AccessKit role is `Role::Unknown`, or if the base AT-SPI role would otherwise translate to the equivalent of "unknown". By always returning `AtSpiRole::Extended`, we deny ATs the opportunity to make inferences based on the nearest standard role, e.g. if the role description is "switch" but the base role is check box. But if returning `AtspiRole::Extended` is the only way to get Orca to pay attention to the role description, then I guess that's a deficiency of AT-SPI.

Also, I believe we need to separate the documentation into its own PR, to avoid making a semver-breaking release of the base accesskit crate. Yes, it's annoying that we have to work around the release automation again.

DataTriny:
@mwcampbell It depends on which property is more important: the end-user might get more information from the localized role description.

"role description" strongly suggests that it can be used in conjunction with platform-specific roles, halas this is not the case with AT-SPI. I chose to do it like this to have a consistent behavior on all platforms, but your arguments are valid as well.

Yes, I'll extract the documentation update once we settle on a resolution.

mwcampbell:
OK, if the only way to get AT-SPI clients to pay attention to the role description is to use `AtspiRole::Extended`, then let's go ahead with that solution.

mwcampbell:
I wonder if we can get this one finished before I publish releases. I'm OK with the current implementation for AT-SPI, using the `Extended` role. We just need to split the documentation change into a separate PR so as not to trigger a semver-breaking release of the common crate.

DataTriny:
Sure! I'll try to do that today or tomorrow.

PR #321, opened 2023-12-14
fix: Bump async-channel dependency to `2.1.1`

mwcampbell:
This is really just a dummy commit because the previous one, having been tagged with `chore`, won't get release-please to release a new version of the unix adapter.


DataTriny (approved):
I messed up here...

PR #322, opened 2023-12-14
fix!: Force a semver-breaking release for the winit rwh feature additions

mwcampbell:
I have to do a dummy commit for the winit change as well, because adding a new default feature that is required for the crate to function is a breaking change.


DataTriny (approved):
This one was on purpose as I didn't think introducing a feature would be a breaking change, especially since the default feature don't change the behavior. But admitedly I didn't seek for advice on this.

PR #324, opened 2023-12-26
fix!: Lazily activate Unix adapters

### LINKED ISSUE #314: AT-SPI2: Activate tree updates only if org.a11y.Status's IsEnabled is true
In Slint we've had the situation that on Linux machines where there's no screen reader running, access kit would still send tree updates on the a11y bus. We've seen this in https://github.com/slint-ui/slint/issues/3867 as well as in https://github.com/slint-ui/slint/pull/3046#issuecomment-1623831111 .

Based on my reading of https://www.freedesktop.org/wiki/Accessibility/AT-SPI2/ as well as peeking at the [Qt Sources](https://codebrowser.dev/qt6/qtbase/src/gui/accessible/linux/dbusconnection.cpp.html#90) as well as [Orca](https://github.com/GNOME/orca/blob/66cd6c8f85e57912a4873bba8d72683b34e12670/src/orca/settings_manager.py#L375) it seems like to me that AccessKit's AT-SPI2 adapter should consider itself active only if `org.a11y.Status.IsEnabled` in the path `/org/a11y/bus` on `org.a11y.Bus` on the session(!) bus is true, and of course watch the property for any changes.

Would this behaviour make sense to implement?

DataTriny:
Fixes #314 

When a Unix adapter is created, we now start monitoring the D-Bus session bus for property change signals, which indicate if assistive technologies are currently running. If this becomes true, all currently registered adapters will be initialized. If for some reason the AT-SPI bus disappear, tree updates will still happen but nothing will be sent through D-Bus. If the AT-SPI bus is launched again, the connection will be re-established, the D-Bus object server will be populated and signals will be emitted like before.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,6 +67,7 @@ dependencies = [
  "async-once-cell",
  "atspi",
  "futures-lite",
+ "futures-util",
  "once_cell",
  "serde",
  "tokio",
@@ -860,6 +861,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "futures-macro"
+version = "0.3.27"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3eb14ed937631bd8b8b8977f2c198443447a8355b6e3ca599f38c975e5a963b6"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
 name = "futures-sink"
 version = "0.3.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -879,6 +891,7 @@ checksum = "3ef6b17e481503ec85211fed8f39d1970f128935ca1f814cd32ac4a6842e84ab"
 dependencies = [
  "futures-core",
  "futures-io",
+ "futures-macro",
  "futures-sink",
  "futures-task",
  "memchr",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -63,6 +63,7 @@ dependencies = [
  "accesskit",
  "accesskit_consumer",
  "async-channel 2.1.1",
+ "async-lock",
  "async-once-cell",
  "atspi",
  "futures-lite",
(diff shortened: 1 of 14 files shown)


mwcampbell:
I just did a first pass through the changes, and the biggest problem that I see is that the implementation of `update` in the winit adapter is incorrect. The `update` method (as opposed to `update_if_active`) is supposed to force the adapter to fully initialize, calling the original tree source function to return the initial tree state, then update the newly created tree with the tree update passed to the `update` method. But the new Unix backend in the winit adapter does nothing. This is incorrect, because when the adapter is eventually lazily initialized, it'll be initialized with the original tree source function; the new update will be lost.

I know that doing a correct implementation of `update` would be complicated in the new implementation of lazy initialization. So I propose that, since we're making a semver-breaking change anyway, we drop the `update` method from the winit adapter entirely. I used to use that method as a shortcut in our sample code, but I don't anymore, and no production-quality GUI toolkit should use it; they should always use `update_if_active`. To make sure that API-breaking change is prominent in the changelog, I suggest doing that change in its own PR, then rebasing this one.

DataTriny:
I wanted to get rid of it as well. I'll make the changes.

mwcampbell:
Likewise in the C SDL example, we should drop `accesskit_sdl_adapter_update`, which the example doesn't actually use.

mwcampbell:
What happens if there's no D-Bus session bus running? It looks to me like the program would crash, though I haven't yet verified this. While a full-featured desktop environment will always have a D-Bus session bus, some developers will want their GUI applications to run in a stripped-down environment without D-Bus, while also supporting accessibility for those who need it. So can we handle the lack of a session bus more gracefully?

DataTriny:
My assumption was that the session bus would indeed always be there. I'm not even sure if it exists a desktop environment that doesn't require D-Bus. I'll see what I can do though.

mwcampbell:
I wonder if @federicomenaquintero and @zeenix would have time to review this PR, to ensure it's implementing AT-SPI correctly and making the best possible use of zbus and the async Rust ecosystem. They might notice things that aren't obvious to me.

mwcampbell:
@DataTriny Based on the results of your testing, I'm willing to go ahead and merge this without further review from Federico or Zeeshan. If any bugs or suboptimal implementations are found, those can be fixed later. I'm satisfied that this PR doesn't introduce any regression from what we already have.

zeenix:
> I'm willing to go ahead and merge this without further review from Federico or Zeeshan.

I'm terribly sorry but I just didn't have time. :( I did have a very quick look and nothing looked bad.

mwcampbell:
@zeenix Thanks for giving it a quick look.

PR #325, opened 2023-12-27
fix!: Remove `accesskit_winit::Adapter::update`

DataTriny:
As discussed in #324, the `update` method on the winit `Adapter` isn't really used anymore. Since upcoming changes to the Unix adapter will render this method useless on this platform, this PR removes it.

`update_if_active` is now the only way to push tree updates.

Note that I haven't removed `update` methods from the various subclassing adapters, but they might be useless as well now.


mwcampbell:
I'm OK with leaving the Windows and macOS subclassing adapters alone for now.

PR #334, opened 2024-01-06
fix: Decrease minimum Python version to 3.8 for Python bindings

DataTriny:
All these changes are to better align with [kivy](https://github.com/kivy/kivy)'s build matrix. The only missing target is ARMv6L, but it isn't supported by Rust.

We could also build for apple-universal2 instead of x86_64 and aarch64 separately. Cargo doesn't support this but maturin has a special post-linking step for it.

Given the nature of the changes, I don't think accesskit_python's semver should be increased.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.9'
+          python-version: '3.8'
 
       - name: restore cache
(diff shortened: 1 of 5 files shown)


mwcampbell:
Adding a macOS universal wheel does sound like a good idea. How hard is that to do?

Also, any reason not to use the "fix" tag for this PR? "chore" won't trigger a release.

DataTriny:
I've tested on my fork that this change to the CI is enough to build an universal2 wheel. The test step shows that pip can install it just fine too.

PR #336, opened 2024-01-08
fix: Make full use of the tokio ecosystem if the `tokio` feature is enabled on Unix

DataTriny:
I noticed that we can get rid of a couple of async-related crates when the `tokio` feature is enabled, because tokio ships its own async types.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,6 +83,7 @@ dependencies = [
  "once_cell",
  "serde",
  "tokio",
+ "tokio-stream",
  "zbus",
 ]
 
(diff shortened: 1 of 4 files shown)


mwcampbell:
I noticed that the following line was removed in the select loop:

```
complete => return Ok(()),
```

I'm not sure I understand what that line did. Was it not needed in the first place, or is there some other change that makes it unnecessary even for async-io?

DataTriny:
In practice this line is never executed since the app will terminate before both streams close. Since tokio's version of select doesn't allow this "compile" keyword, I decided to remove it. I think we could put it back if you want, with `#[cfg(not(feature = "tokio))]`.

mwcampbell:
If it's never needed in practice, I'm fine with leaving it out.

PR #346, opened 2024-01-18
fix: Avoid `unnecessary_literal_unwrap` clippy lint

mwcampbell:
This prevents clippy from being the limiting factor in our MSRV. The MSRV for `accesskit_windows` is now 1.64.0 with or without clippy.


DataTriny (approved):
I didn't want to do this but, why not.

PR #348, opened 2024-01-18
ci: Enforce MSRV

DataTriny:
Blocked on #344 

Ensures that AccessKit can compile with the minimum Rust version as advertised in the workspace's `Cargo.toml`.

`cargo clippy` deliberately runs on stable so we can allow/deny lints that are not present on older Rust versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,19 +42,18 @@ jobs:
       - uses: actions/checkout@v3
       - uses: EmbarkStudios/cargo-deny-action@v1
 
-  test:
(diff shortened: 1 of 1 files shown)


mwcampbell:
This looks fine, but it should probably be rebased before merging.

mwcampbell:
Just a reminder that this looks fine, but to make sure we do a clean merge, I believe this branch needs to be rebased.

DataTriny:
@mwcampbell You might have missed the notification, it was rebased days ago. I'll merge this now.

PR #353, opened 2024-02-20
chore: Drop release-please

mwcampbell:
I've grown dissatisfied with the inflexibility of release-please. The particular scenario that has finally led me to seriously consider dropping it is this: When we refactor accesskit_consumer to support full copy-on-write tree snapshots again via immutable-chunkmap, I don't want this change to cause semver-breaking version increments in the downstream crates (the platform adapters). But with release-please, that would happen automatically. So I think, unless we eventually find better tools for release automation, we'll have to go back to a manual release process. @DataTriny What do you think?


mwcampbell:
I'm abandoning this PR, because I think we decided that we'll just keep working around the quirks of release-please.

PR #356, opened 2024-02-24
fix: Use the new accesskit_atspi_common crate in the Unix adapter

mwcampbell:
as discussed in #352


DataTriny (approved):
Thank you for dusting off the whole crate. I wish I wrote some of that myself.

PR #357, opened 2024-02-24
fix: Add missing README

mwcampbell:
I realized that I can't actually publish `accesskit_atspi_common` to crates.io because I didn't add the README.


DataTriny:
It also broke the Python package.

DataTriny:
@mwcampbell Can you please try to generate a source distribution of the Python bindings by running `maturin sdist` and see if it succeeds. It should now, but just to be safe. Thanks.

mwcampbell:
It works now.

DataTriny (approved):
Thanks for checking.

PR #359, opened 2024-02-25
fix: Bump windows-rs dependency to 0.52

### LINKED ISSUE #350: Update `windows` crate
https://github.com/AccessKit/accesskit/blob/main/platforms/windows/Cargo.toml is stuck on 0.48

It would be nice if a range of versions could be supported, because that would make it less likely that users will get duplicates of `windows` (which is a huge crate that compiles very slowly)

DataTriny:
Closes #350 

I don't think we can do more than that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2378,35 +2378,45 @@ checksum = "712e227841d057c1ee1cd2fb22fa7e5a5461ae8e48fa2ca79ec42cfc1931183f"
 
 [[package]]
 name = "windows"
-version = "0.48.0"
+version = "0.52.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e686886bc078bc1b0b600cac0147aadb815089b6e4da64016cbd754b6342700f"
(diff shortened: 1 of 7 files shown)


mwcampbell:
I had actually done the same thing on the windows-0.52 branch several weeks ago (before the issue was opened), but then stopped because I wasn't sure how to handle the calls to `as_impl` in text.rs, which now have to be marked `unsafe`. But it looks like `as_impl` was actually always unsafe anyway, so this isn't really a regression. I guess it's actually fine, because we're casting the `ITextRangeProvider` interface to our implementation, and because of how UIA works, we know that the interface pointer comes from the same application process, and it's highly unlikely that there are two full UIA providers in the same process. So I'll go ahead and merge this and delete my duplicate branch.

mwcampbell:
Oh, sorry, it looks like I never actually pushed my windows-0.52 branch. So you couldn't have known that I had already done this. Thanks for doing it.

DataTriny:
@mwcampbell I came to the same conclusion about calls to `as_impl`, I should probably have mentioned it in the PR description. Thanks for clarifying this then!

PR #366, opened 2024-03-06
refactor: Use `#[default]` attribute macro on `Role`

DataTriny:
This just makes it a bit more obvious which `Role` is the default, as it's now written near the actual variant.


DataTriny:
Woops, looks like we have a security update to apply.

PR #367, opened 2024-03-06
fix: Update mio dependency to address RUSTSEC-2024-0019

DataTriny:
This doesn't actually affect us since the vulnerability is Windows-specific, and we only rely on tokio on *nix platforms.

See the advisory for [RUSTSEC-2024-0019](https://rustsec.org/advisories/RUSTSEC-2024-0019).


DataTriny:
Given the security implications of this PR, I'm going to merge it now.

PR #370, opened 2024-03-11
fix: Expose root node ID in `accesskit_atspi_common::Adapter`

mwcampbell:
I need this in my `newton_atspi_compat` crate so I can create a `PlatformNode` from the window root outside of the AT-SPI event handler._


mwcampbell:
Merging this now, because it's a trivial method that's making existing functionality public, and I need it for something I'm working on now.

PR #373, opened 2024-03-16
fix: Update `windows` to `0.54`

DataTriny:
The major highlight of this PR is that `VariantFactory` is replaced by a simpler `Variant` newtype. The `VARIANT` type from the `windows` crate is used directly whenever possible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2377,9 +2377,9 @@ checksum = "712e227841d057c1ee1cd2fb22fa7e5a5461ae8e48fa2ca79ec42cfc1931183f"
 
 [[package]]
 name = "windows"
-version = "0.52.0"
+version = "0.54.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


mwcampbell:
@DataTriny Are you sure that calling `HRESULT::from_win32` on `UIA_E_ELEMENTNOTAVAILABLE` and `UIA_E_INVALIDOPERATION` is correct? Those constants are already `HRESULT` codes, not Win32 errors. I guess we need tests that cover those errors.

mwcampbell:
I think what we need to do is just directly construct the `HRESULT` struct, like this:

```
HRESULT(UIA_E_ELEMENTNOTAVAILABLE as _)
```

This is based on looking at how constants like `windows::Win32::Foundation::E_NOTIMPL` are written.

DataTriny:
@mwcampbell I tried printing a debug representation of those errors and I think you're right.

DataTriny:
I previously made sure that `Error::empty()` was equivalent to `Error::OK`.

PR #375, opened 2024-03-23
refactor!: New approach to lazy initialization

### LINKED ISSUE #295: Moving lazy tree initialization back into the core platform adapters
I now believe that #179 was a mistake. The issue that motivated that PR, with the way I was doing lazy initialization before, was real. But I think I need to find another way to fix it. On Unix, an optimal implementation of AT-SPI is going to require that the Unix platform adapter directly support lazy initialization of the tree _after_ the adapter has been instantiated. And even on Windows and macOS, this is ideally something we should delegate to the core adapters themselves, even for non-subclassing use cases. And as I noted in #179, on Windows, the UIA initialization marker is a wart in the API. Finally, I noticed just last night that I had accidentally made Windows initialization less lazy while working on #179, because `handle_wm_getobject`'s behavior of aborting for some object IDs is only effective if that method itself does the lazy initialization.

I think what I need to do is combine the lazy tree source and the action handler into a single trait object, rather than using a generic closure for the tree source. I'll experiment and open a PR once I'm reasonably sure I've got the right design this time.

mwcampbell:
This fixes #295 while still keeping the advantages of #179 for clean integrations on Windows and macOS.

The placeholder tree, if necessary, is now the responsibility of the platform adapter; that makes the adapter somewhat more complicated, but it's better to have that kind of complexity in AccessKit than in the caller. The AT-SPI implementation doesn't need a placeholder tree, and neither will Newton (the new Wayland-native accessibility stack I'm prototyping). I don't like how `accesskit_atspi_common` has to expose some of its implementation details to enable robust state transitions with minimal overhead in `accesskit_unix`, but I don't know a better way, and `accesskit_atspi_common` is basically an internal crate anyway.

The update to egui is in [this branch](https://github.com/mwcampbell/egui/tree/accesskit-new-lazy-init). The changes are mostly in `egui-winit` and `eframe`.

@DataTriny Please repeat the same testing you did when you implemented lazy initialization for AT-SPI.

I don't yet know how to get all the benefits of #179, particularly synchronous tree initialization without requiring the application/toolkit state to be wrapped in a reference-counted mutex, in the Unix adapter. The problem is that on Windows and macOS, tree initialization is always done in response to a platform event which, in a clean integration, is initially handled by the toolkit, so it can pass us a short-lived mutable reference to the activation handler. But in the Unix adapter, we listen to accessibility status changes on our own thread. Maybe `accesskit_unix` could have an alternate API where the application listens to that property itself and calls a function when it changes, passing us a short-lived reference to the activation handler.


DataTriny:
The behavior of lazy initialization on Unix seem to still work as expected. Reviewing the code now.

DataTriny (approved):

PR #376, opened 2024-03-23
fix: Fix new compiler warning in Rust 1.77

mwcampbell:
The compiler is now issuing a warning about taking a reference to a mutable static; this will become a hard error in the 2024 edition.

### DIFF
--- a/platforms/macos/src/subclass.rs
+++ b/platforms/macos/src/subclass.rs
@@ -33,7 +33,7 @@ static SUBCLASSES: SyncLazy<Mutex<HashMap<&'static AnyClass, &'static AnyClass>>
 static mut ASSOCIATED_OBJECT_KEY: u8 = 0;
 
 fn associated_object_key() -> *const c_void {
-    unsafe { &ASSOCIATED_OBJECT_KEY as *const u8 as *const _ }
+    unsafe { std::ptr::addr_of_mut!(ASSOCIATED_OBJECT_KEY) as *const u8 as *const _ }
 }
 
 type LazyAdapter = Lazy<Adapter, Box<dyn FnOnce() -> Adapter>>;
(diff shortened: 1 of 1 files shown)


platforms/macos/src/subclass.rs
```diff
@@ -33,7 +33,7 @@ static SUBCLASSES: SyncLazy<Mutex<HashMap<&'static AnyClass, &'static AnyClass>>
 static mut ASSOCIATED_OBJECT_KEY: u8 = 0;
 
 fn associated_object_key() -> *const c_void {
-    unsafe { &ASSOCIATED_OBJECT_KEY as *const u8 as *const _ }
+    unsafe { std::ptr::addr_of_mut!(ASSOCIATED_OBJECT_KEY) as *const u8 as *const _ }
```
kpreid, line 36: The `unsafe` block can be removed entirely, and the cast replaced with stricter `pointer::cast`:

```suggestion
    std::ptr::addr_of!(ASSOCIATED_OBJECT_KEY).cast::<u8>();
```

mwcampbell: I tried your suggestion and got this compiler error:

error[E0133]: use of mutable static is unsafe and requires unsafe block

DataTriny:
I'm not even sure to understand why `ASSOCIATED_OBJECT_KEY` need to be `mut` in the first place. How could there be more than one? It is declared as static, not as const, so it refers to a value written somewhere in the resulting binary.

mwcampbell:
I'm assuming that a sufficiently smart compiler or linker could deduplicate multiple non-mutable statics into the same address, or even turn a non-mutable static into a constant with no global address at all.

DataTriny:
If we can rely on [this page of the Rust Reference book](https://doc.rust-lang.org/reference/items/static-items.html), I think what you describe would be considered a bug as what I said before seem to be a guarantee upholed by the compiler.

mwcampbell:
OK, I'm convinced. I wrote a simple program that declares three immutable statics, all with the same value (0), and prints their addresses. In a fully optimized release build on macOS, for both ARM64 and x86-64, they have distinct addresses.

DataTriny:
Do we still need to call `addr_of` then? The warning only applied to mutable references right?

DataTriny (approved):

platforms/macos/src/subclass.rs
kpreid: Sorry for the incorrect information — it seems I incorrectly assumed that since no _reference_ was being taken, `addr_of!` would be safe.

PR #379, opened 2024-04-04
fix: Set INTERFACE_INCLUDE_DIRECTORIES in CMake config

mwcampbell:
The Meson build system needs this in order to find the header file. I encountered this problem while working on integrating AccessKit into GTK. Our current build system may not work for distro packagers, but this fix at least lets me make progress on that project, and may also help other users.

### DIFF
--- a/bindings/c/README.md
+++ b/bindings/c/README.md
@@ -16,12 +16,6 @@ You can then require AccessKit as a dependency by adding this to your `CMakeList
 find_package(ACCESSKIT REQUIRED)
 ```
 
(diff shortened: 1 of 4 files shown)


DataTriny (commented):
As a result it looks like calling `include_directories` inside the build process of our examples is not necessary anymore. Can you please remove these?

The main README will also need updating.

mwcampbell:
@DataTriny Thanks for adding the changes you suggested. Looks like this is ready to merge.

PR #382, opened 2024-04-21
feat: Implement the `description` property

DataTriny:
I know at least Slint implements it.

There is apparently no corresponding event to fire on macOS. Also on this platform: Chromium includes it in live region updates, but we don't do that on other platforms so this must be a Chromium specific thing.

Tested on all platforms by modifying the winit example like so:

```rust
fn build_button(id: NodeId, name: &str, classes: &mut NodeClassSet) -> Node {
    let rect = match id {
        BUTTON_1_ID => BUTTON_1_RECT,
        BUTTON_2_ID => BUTTON_2_RECT,
        _ => unreachable!(),
    };

    let mut builder = NodeBuilder::new(Role::Button);
    builder.set_bounds(rect);
    builder.set_name(name);
    builder.set_description("Announces that this button was pressed.");
    builder.add_action(Action::Focus);
    builder.set_default_action_verb(DefaultActionVerb::Click);
    builder.build(classes)
}
```

### DIFF
--- a/consumer/src/node.rs
+++ b/consumer/src/node.rs
@@ -51,6 +51,7 @@ impl<'a> Node<'a> {
             is_focused: self.is_focused(),
             is_root: self.is_root(),
             name: self.name(),
+            description: self.description(),
             value: self.value(),
             live: self.live(),
             supports_text_ranges: self.supports_text_ranges(),
@@ -541,6 +542,12 @@ impl<'a> Node<'a> {
         }
     }
 
+    pub fn description(&self) -> Option<String> {
+        self.data()
+            .description()
+            .map(|description| description.to_string())
+    }
+
     pub fn value(&self) -> Option<String> {
         if let Some(value) = &self.data().value() {
             Some(value.to_string())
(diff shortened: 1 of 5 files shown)


DataTriny:
Oh no... This is a breaking change on Unix since I changed the signature of `PlatformNode::description`. Need to split this PR then.

mwcampbell:
Actually, I don't think the Unix implementation needs to be split into its own PR, because `NodeWrapper` is private to `accesskit_atspi_common`, even though the methods of `NodeWrapper` were marked `pub` rather than `pub(crate)`.

DataTriny:
@mwcampbell You're right! I'll fix the remaining incorrect visibility qualifiers next then.

mwcampbell:
The `Node::description` method in `accesskit_consumer` ought to have support for the `described_by` property, just as `Node::name` supports `labelled_by`. But unlike `Node::name`, I don't think `Node::description` should look at descendant nodes.

Apart from that, this looks fine.

DataTriny:
I didn't do it because I'm not a fan of how we currently implement the labelled by relationship. It make sense on platforms where there isn't a possibility to expose this relationship to ATs but this shouldn't be the default IMO.

I think there is a bit of work to do on the atspi crate to improve the `RelationSet` type before we can do this cleanly though.

PR #384, opened 2024-04-23
Use new objc2 crates

madsmtm:
`icrate` has been split into multiple smaller crates, and the feature setup in that has been reworked, which should improve compilation times.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -64,8 +64,9 @@ version = "0.12.0"
 dependencies = [
  "accesskit",
  "accesskit_consumer",
- "icrate 0.1.0",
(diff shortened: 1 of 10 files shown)


DataTriny (approved):
Thanks @madsmtm, looks good!

I think we'll want to integrate this when we update to winit 1.30 though.

See https://github.com/rust-windowing/winit/issues/3476

mwcampbell:
I see no reason to delay merging this. After all, our objc2 dependencies are already ahead of winit 0.29; this PR doesn't make it any worse. @DataTriny Do you disagree?

PR #385, opened 2024-04-23
feat: Expose the class name property

mwcampbell:
As far as I know, only UIA on Windows has such a property. But this simple feature is helping me debug my GTK AccessKit integration, since it makes it easy to find the GTK class of any accessible object, e.g. via NVDA+F1.

### DIFF
--- a/consumer/src/node.rs
+++ b/consumer/src/node.rs
@@ -622,6 +622,10 @@ impl NodeState {
     pub fn raw_value(&self) -> Option<&str> {
         self.data().value()
     }
+
+    pub fn class_name(&self) -> Option<&str> {
+        self.data().class_name()
+    }
 }
 
 impl<'a> Node<'a> {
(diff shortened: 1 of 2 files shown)


DataTriny:
I can easily see how this can be useful when debugging. Apart from this very specific usecase reasons why ATs would need that aren't very appealing to me. I'm pretty sure NVDA relies on it to patch some broken widget implementations, but I don't think AccessKit should encourage this way of dealing with issues. I think this information is too deep in UI toolkits internals to be exposed. Furthermore with other programming languages UI toolkit developers will have to resort to expensive reflection to provide this.
My opinion would be to entirely drop the property from our schema.

mwcampbell:
We always have to be pragmatic about things like this. The goal is as much accessibility as possible, by any means necessary. That includes workarounds implemented by the screen reader. If a screen reader workaround based on a widget's class name could help a blind person get or keep a job, then it's worth it. Of course, it won't be practical for all toolkits to expose this property, but we should make the option available.

DataTriny:
I think the bug report use case would be better served by having a good implementation of the `Debug` trait for `Node`, something I'm interested in tackling.

But I don't want to block this any longer. Implementors who can easily provide this are free to do so, even if I still think it is not a good idea.

PR #387, opened 2024-04-24
feat: Add new `AuthorControlledParent` role

mwcampbell:
This role, and the associated logic in `accesskit_consumer`, is meant to handle a pattern I'm seeing in GTK, and which is probably common in other object-oriented toolkits with a one-to-one correspondence between widgets and accessibility nodes. In GTK, there are some widget classes like `GtkMenuButton` that are essentially wrappers around a more generic widget like `GtkToggleButton`. In these cases, the more generic widget is the one that's actually focusable, and the wrapper ought to be filtered out of the platform accessibility tree. But the wrapper is the one that's actually created by the application developer, and is the place where they can set properties like the accessibility name and description. GTK's accessibility implementation currently has special-case logic for pulling the name and description from the parent for these specific widgets, but I think this is a general pattern that's worth codifying in AccessKit. I wrote test cases to verify that the logic I had in mind is actually implemented. This does make `accesskit_consumer::Node::labelled_by` more complicated, but in general, I think we should be willing to make the consumer more complicated in order to simplify toolkit implementations.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -113,6 +113,16 @@ pub enum Role {
     Application,
     Article,
     Audio,
+    /// This role is for cases where a specialized widget, such as a menu
+    /// button, merely serves as a parent for a more generic widget, such as
+    /// a toggle button, which is the actual focusable widget, but the parent
+    /// widget is the one where the application developer sets the essential
+    /// name and description properties, or their corresponding relations.
+    /// When this role is used, this node is filtered out of the platform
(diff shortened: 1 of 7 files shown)


mwcampbell:
I'm abandoning this PR. One of my guiding questions for AccessKit now is, "Am I sure I want to ship this in 1.0?" And for this PR, the answer to that question is no. This PR was prompted by a problem in GTK, so I'll do a workaround in my GTK AccessKit implementation instead.

PR #388, opened 2024-04-25
refactor!: Rename `Checked` to `Toggled`; drop `ToggleButton` role

mwcampbell:
ARIA and AT-SPI have both "checked" and "pressed" flags, but they're functionally equivalent; the only difference is that one is for checkboxes while the other is for toggle buttons. I prefer having just one property. AccessKit already had only a "checked" property, which the AT-SPI adapter translated to "pressed" if appropriate. But I think it makes sense to rename our "checked" property to something more generic, since "checked" only makes sense for checkboxes. Here I took some inspiration from UIA, though I went with the more succinct "toggled" rather than UIA's `ToggleState`.

I also think it makes sense to not have a dedicated `ToggleButton` role. Instead, now a `Button` becomes a toggle button if the `toggled` property is set. This is the same way that buttons become toggle buttons in ARIA.

If it seems that I'm being inconsistent about whether to add dedicated roles (as I did in a previous refactor) or use state flags to modify the role (as I'm doing here), the guiding principle is this: eliminate as many invalid combinations of role and state as we can. When I previously eliminated state flags and added more roles instead, the only purpose of the state flag was to modify one specific role. But here, where we have a state property that can be false, true, or unset (unlike the simple flags I previously eliminated), the property serves to both modify the role (two different roles in this case) and indicate some state. The goal is the same, by eliminating the `ToggleButton` role, we make it impossible to mess up the accessibility node by having a `ToggleButton` with no toggle state, and now, having a plain `Button` with toggle state is the one way to do a toggle button.


bindings/c/cbindgen.toml
```diff
@@ -35,7 +35,7 @@ renaming_overrides_prefixing = true
 "Affine" = "accesskit_affine"
 "AriaCurrent" = "accesskit_aria_current"
 "AutoComplete" = "accesskit_auto_complete"
-"Checked" = "accesskit_checked"
+"Toggled" = "accesskit_toggled"
```
DataTriny, line 38: Can you move this line down so we keep the list in alphabetical order?

bindings/c/cbindgen.toml
```diff
@@ -59,7 +59,7 @@ renaming_overrides_prefixing = true
 "WPARAM" = "WPARAM"
 "opt_AriaCurrent" = "accesskit_opt_aria_current"
 "opt_AutoComplete" = "accesskit_opt_auto_complete"
-"opt_Checked" = "accesskit_opt_checked"
+"opt_Toggled" = "accesskit_opt_toggled"
```
DataTriny, line 62: Can you move down this one too?

DataTriny (changes requested):
I totally agree. Someone will have to revert https://github.com/slint-ui/slint/pull/5033 but that's fine.

DataTriny (approved):

PR #392, opened 2024-04-27
feat: Add the `owns` relation

mwcampbell:
This is a standard ARIA relation, so I'm surprised we didn't already inherit it from Chromium; I doubt I would have deliberately left it out or removed it. Anyway, I noticed it while going through the relations in GTK.


DataTriny:
Is it a good idea to add it without mentioning right away that the `children` property is preferred? As a non-native English speaker it isn't obvious to me that the `owns` property actually stores direct descendants.

mwcampbell:
Good point. Hopefully the documentation I just added is clear enough.

DataTriny:
It is. Thanks!

PR #393, opened 2024-04-27
refactor!: Clean up table roles and properties

mwcampbell:
The roles and properties that I copied from Chromium are complicated compared to the ARIA standard, and the meanings of the various properties were never documented. I think that if the ARIA roles, properties, and (lack of) relations are good enough for both GTK 4 and ARIA itself, they're probably good enough for us, too. I'm quite sure we don't need the `Column` role; it doesn't even have a proper mapping on two of the three platforms. As for the old `table_header`, `table_row_header`, and `table_column_header` relations, I guess the AT (or in our case, maybe the consumer library) should scan for the headers rather than requiring the toolkit to specify such relations; there are certainly no such relations in ARIA.

The new `row_index_text` and `column_index_text` properties are based on ARIA properties that are also in GTK 4, that we didn't previously have.


mwcampbell:
Oh, and I dropped the `table_` prefix from the property names because, as with `hierarchical_level`, long names are just more cumbersome.

DataTriny (approved):

DataTriny:
I fixed merge conflicts from #390. I'll merge now.

PR #394, opened 2024-04-28
clippy: Fix new legacy-numeric-constants lint.

waywardmonkeys:
It is preferred to use the associated constants as the other forms are expected to be deprecated in the future.


DataTriny (approved):
Thanks!

PR #397, opened 2024-04-29
fix!: Update winit to 0.30

DataTriny:
Examples now use the new `ApplicationHandler` trait, which led to significant changes.

This is a breaking change to `accesskit_winit`, as described in the [changelog](https://docs.rs/winit/latest/winit/changelog/index.html).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -64,7 +64,7 @@ version = "0.12.0"
 dependencies = [
  "accesskit",
  "accesskit_consumer",
- "objc2 0.5.1",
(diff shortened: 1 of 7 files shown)


mwcampbell:
In the examples, I don't see any reason why the `Windo` needs to be wrapped in an `Arc`. The `Arc` is never cloned, only borrowed.

Aside from that, this looks good.

DataTriny:
That was a leftover from the official winit example. Fixed now.

Issue #400, opened 2024-05-01
accesskit_c: v0.9.0 - missing and mixed up macOS library architectures


bruvzg:
The latest [accesskit_c-v0.9.0](https://github.com/AccessKit/accesskit/releases/tag/accesskit_c-v0.9.0) release has only `macos/arm64` subfolder (`macos/x86_64` is missing), which contains `x86-64` libraries (both shared and static) instead of `arm64`.

bruvzg:
https://github.com/AccessKit/accesskit/blob/e771ca38850d797c35c5679229007e8c6cb134e1/.github/workflows/c-bindings.yml#L14-L20

This might be related to GitHub recently starting to switch `macOS-latest` runners from x86 to arm.

PR #406, opened 2024-05-05
fix: Explicitly set macOS CI build architecture for C bindings.

bruvzg:
Fixes https://github.com/AccessKit/accesskit/issues/400

`macos-latest` GitHub CI runners are now [using `arm64` image](https://github.com/actions/runner-images/blob/macos-14-arm64/20240422.3/README.md), and cmake for C bindings [determine target folder based on the current system architecture](https://github.com/AccessKit/accesskit/blob/09d91577dd88743e379a1fdea34b25a94726d0fb/bindings/c/accesskit.cmake#L31-L39) if it's not set explicitly, and `x86_64` build was overwriting `arm64 `build artifacts:


DataTriny (approved):
Thanks @bruvzg!

PR #408, opened 2024-05-13
fix: Fix the filtered sibling iterators to use the filtered parent to find the back node

mwcampbell:
This was broken when calling `preceding_filtered_siblings` or `following_filtered_siblings` on the only child of an ignored parent. I extended the tests to cover that case.


DataTriny (approved):
Good catch!

PR #416, opened 2024-05-19
fix: Clamp character indices when converting a text selection to a range

mwcampbell:
@DataTriny discovered a way to cause a panic in AccessKit when moving in an egui multi-line edit. Apparently, the egui cursor can end up being outside the actual bounds of the text. I could fix this in egui, but I think it's better to make AccessKit resilient to broken implementations, when we can do so at little cost. So I implemented a fix in AccessKit instead.

To test this fix with egui, you'll need to cherry-pick this PR onto a local copy of the winit-0.29-backport branch.

### DIFF
--- a/consumer/src/text.rs
+++ b/consumer/src/text.rs
@@ -32,6 +32,20 @@ impl<'a> InnerPosition<'a> {
         })
     }
 
+    fn clamped_upgrade(tree_state: &'a TreeState, weak: WeakPosition) -> Option<Self> {
+        let node = tree_state.node_by_id(weak.node)?;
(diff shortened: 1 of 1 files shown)


DataTriny:
For reference, here are the steps that would trigger a panic:

- Write some text inside an egui multi-line text area,
- Add a blank line at the end,
- Go back to the line above (up arrow),
- Move the caret to the right by at least one character,
- Now go down to land on the last blank line.

I can confirm that this PR fixes the bug.

DataTriny (approved):

PR #418, opened 2024-05-20
chore: Update immutable-chunkmap

mwcampbell:
My immutable-chunkmap PR to replace the dependency on packed_struct with a hard-coded equivalent is merged, and version 2.0.5 was released. So now we can update immutable-chunkmap and eliminate several transitive dependencies.

Note: This PR by itself won't trigger any releases, but we have releases pending anyway, including our precompiled artifacts (the C and Python bindings), which are the only releases where an update to our Cargo.lock actually matters.


mwcampbell:
Actually, it occurred to me that it would be better to update the minimum version of the crate that accesskit_consumer depends on. Let me do that and open a new PR with the correct conventional commit tag.

PR #419, opened 2024-05-20
fix: Update minimum version of immutable-chunkmap

mwcampbell:
accesskit_consumer now requires at least immutable-chunkmap 2.0.5 because that's the first version that doesn't depend on packed_struct and all of its transitive dependencies.


DataTriny (approved):
Congratulations on that!

PR #420, opened 2024-05-22
fix: Improve how coordinates are computed on Unix

DataTriny:
Orca apparently changed the way it queries for bounds. Now only the client area of a window is considered. I chose to not drop the `WindowBounds::outer` field because we might need it again later, and it is public.

We also now support querying coordinates in relation to a parent, a feature that I haven't seen used but could be useful when dealing with text for instance.


DataTriny:
Good point: my initial implementation was based on tests with GTK3 as far as I remember. But in GTK4 the coordinates space seem to be the direct parent of the Component.

With that being said, neither Chromium nor Orca use it.

DataTriny:
I made more tests and pushed more changes as a result. I'm quite confident in this implementation now.

PR #423, opened 2024-05-27
fix: Fix a logic error in suffix calculation for text changes

mwcampbell:
To reproduce the bug, add two newlines to the end of an empty multi-line edit control. When the second newline is added, the old text will be "\n" and the new text will be "\n\n". Before this fix, the function mistakely calculated 1 byte as a common suffix since both strings ended with a newline, even though that same newline in the old string was also the common prefix.


DataTriny (approved):
Good catch!

PR #426, opened 2024-06-06
winit: Don't treat `accesskit_unix` as a feature

waywardmonkeys:
We don't need to check both for a feature (enabled by default) and the various `target_os` values.

As a result, the `accesskit_unix` dependency is no longer optional on the Unix platforms.

### DIFF
--- a/platforms/winit/Cargo.toml
+++ b/platforms/winit/Cargo.toml
@@ -15,7 +15,7 @@ rust-version.workspace = true
 features = ["winit/rwh_06", "winit/x11", "winit/wayland"]
 
 [features]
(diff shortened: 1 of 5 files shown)


waywardmonkeys:
I'm on macOS, so I couldn't test this locally myself.

DataTriny:
We added this feature because some projects (namely Bevy) didn't want to enable the Unix adapter by default, see #214. Given the fact that the situation didn't change much, I am not willing to reconsider this right now.

Could you please clarify your intention in submitting this PR @waywardmonkeys?

waywardmonkeys:
I was incorrectly fixing something that triggers an error in 2024 edition ... I'll do a correct fix in the future!

PR #429, opened 2024-06-09
refactor!: Transfer ownership of action requests in the C API

mwcampbell:
This is the simplest way to allow C API users, like GTK, to asynchronously handle arbitrary action requests on another thread.

### DIFF
--- a/bindings/c/src/common.rs
+++ b/bindings/c/src/common.rs
@@ -1046,6 +1046,11 @@ impl From<ActionRequest> for action_request {
     }
 }
 
+#[no_mangle]
+pub extern "C" fn accesskit_action_request_free(request: *mut action_request) {
+    drop(unsafe { Box::from_raw(request) });
+}
+
 type ActivationHandlerCallbackUnwrapped = extern "C" fn(userdata: *mut c_void) -> *mut tree_update;
 pub type ActivationHandlerCallback =
     Option<extern "C" fn(userdata: *mut c_void) -> *mut tree_update>;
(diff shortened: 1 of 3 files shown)


bindings/c/src/common.rs
```diff
@@ -1080,9 +1085,9 @@ impl ActivationHandler for FfiActivationHandler {
 }
 
 type ActionHandlerCallbackUnwrapped =
-    extern "C" fn(request: *const action_request, userdata: *mut c_void);
+    extern "C" fn(request: *mut action_request, userdata: *mut c_void);
 pub type ActionHandlerCallback =
```
DataTriny, line 1089: I think we should document the ownership transfer here and mention `accesskit_action_handler_free`.

DataTriny (changes requested):

bindings/c/src/common.rs
```diff
@@ -1080,9 +1085,12 @@ impl ActivationHandler for FfiActivationHandler {
 }
 
 type ActionHandlerCallbackUnwrapped =
-    extern "C" fn(request: *const action_request, userdata: *mut c_void);
+    extern "C" fn(request: *mut action_request, userdata: *mut c_void);
+
+/// Ownership of `request` is transfered to the callback. `request` must
+/// be freed using [`accesskit_action_request_free`].
```
DataTriny, line 1091:
```suggestion
/// be freed using `accesskit_action_request_free`.
```

Remember that these doc comments only end up in the header file and not on docs.rs, so I think the brackets are only adding noise.

DataTriny (approved):

PR #430, opened 2024-06-09
feat: Feature-gate the `accesskit_atspi_common::simplified` module

DataTriny:
Most users don't actually need this, they should not have to compile it. I'm not set on the name of the feature, feel free to suggest something else.


mwcampbell:
I don't think we even need to make simplified-api a default feature. If we remove that feature from the defaults, then we don't need to touch accesskit_unix, and we can avoid the semver-breaking bump to that crate.

PR #432, opened 2024-06-14
fix: Avoid unnecessary repeated checks for text changes

mwcampbell:
This is necessary if a large number of nodes have changed in a large document, but the text itself hasn't changed. That currently happens when scrolling GtkTextView, and while I could probably optimize that specific case somehow, this optimization is easier and, as far as I can tell, doesn't break anything.


DataTriny (approved):
I can't think of any downside to this optimization either. The only early return happens when the text is unchanged.

PR #448, opened 2024-09-05
fix: Extend the implicit labelled-by relation to more parent roles

mwcampbell:
Since we already have an implicit labelled-by relationship for buttons and links, I think it makes sense to extend it to more parent roles where it should obviously apply. I realized the need for this when I tried a Xilem example with NVDA and discovered that while Xilem assumes that the implicit relationship applies to check boxes, it actually didn't.

### DIFF
--- a/consumer/src/node.rs
+++ b/consumer/src/node.rs
@@ -999,53 +999,178 @@ mod tests {
 
     #[test]
     fn name_from_descendant_label() {
-        const BUTTON_LABEL: &str = "Play";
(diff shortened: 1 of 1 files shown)


DataTriny (changes requested):
Makes sense. But it would be wise to unit test this method now I guess.

mwcampbell:
We already had the `name_from_descendant_label` test, so I extended it to test all the supported roles.

DataTriny (approved):
Thank you.

PR #452, opened 2024-09-17
fix: Improve debug representation of `Node` and `NodeBuilder`

DataTriny:
Consider this program:

```rust
use accesskit::{Action, DefaultActionVerb, NodeBuilder, Rect, Role};

fn main() {
    let rect = Rect {
        x0: 0.0,
        y0: 0.0,
        x1: 50.0,
        y1: 200.0,
    };
    let mut builder = NodeBuilder::new(Role::Button);
    builder.set_bounds(rect);
    builder.set_name("Send");
    builder.set_busy();
    builder.add_action(Action::Focus);
    builder.set_default_action_verb(DefaultActionVerb::Click);
    println!("{builder:?}");
    let node = builder.build();
    println!("{node:?}");
}
```

This currently outputs:

```
NodeBuilder { role: Button, actions: 2, flags: 64, properties: PropertiesBuilder { indices: PropertyIndices([84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 1, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 2, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 0, 84, 84]), values: [Rect(Rect { x0: 0.0, y0: 0.0, x1: 50.0, y1: 200.0 }), String("Send"), DefaultActionVerb(Click)] } }
Node { role: Button, actions: 2, flags: 64, properties: Properties { indices: PropertyIndices([84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 1, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 2, 84, 84, 84, 84, 84, 84, 84, 84, 84, 84, 0, 84, 84]), values: [Rect(Rect { x0: 0.0, y0: 0.0, x1: 50.0, y1: 200.0 }), String("Send"), DefaultActionVerb(Click)] } }
```

Without knowing how these structs work under the hood, it is impossible to figure out their state. Yet this can be useful when debugging, I think Bevy might even have an inspector system which take advantage of the debug representation of components.

With this PR, the above program now outputs:

```
NodeBuilder { role: Button, actions: [Focus], is_busy: true, name: Some("Send"), default_action_verb: Some(Click), bounds: Some(Rect { x0: 0.0, y0: 0.0, x1: 50.0, y1: 200.0 }) }
Node { role: Button, actions: [Focus], is_busy: true, name: Some("Send"), default_action_verb: Some(Click), bounds: Some(Rect { x0: 0.0, y0: 0.0, x1: 50.0, y1: 200.0 }) }
```

Because we have a lot of fields, I chose to only show them when their value is different from the default (`false`, `None` or `&[]`).

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -1166,7 +1166,7 @@ macro_rules! vec_property_methods {
 }
 
 macro_rules! slice_properties_debug_method {
-    ($name:ident, $($getter:ident)*) => {
+    ($name:ident, [$($getter:ident,)*]) => {
         fn $name(&self, fmt: &mut fmt::DebugStruct) {
             $(
                 let value = self.$getter();
@@ -1185,16 +1185,16 @@ macro_rules! node_id_vec_property_methods {
             ($id, NodeId, $getter, get_node_id_vec, $setter, set_node_id_vec, $pusher, push_to_node_id_vec, $clearer)
         })*
         impl Node {
-            slice_properties_debug_method! { debug_node_id_vec_properties, $($getter)* }
+            slice_properties_debug_method! { debug_node_id_vec_properties, [$($getter,)*] }
         }
         impl NodeBuilder {
-            slice_properties_debug_method! { debug_node_id_vec_properties, $($getter)* }
+            slice_properties_debug_method! { debug_node_id_vec_properties, [$($getter,)*] }
         }
     }
 }
 
 macro_rules! option_properties_debug_method {
-    ($name:ident, $($getter:ident)*) => {
+    ($name:ident, [$($getter:ident,)*]) => {
         fn $name(&self, fmt: &mut fmt::DebugStruct) {
             $(
                 let value = self.$getter();
@@ -1213,10 +1213,10 @@ macro_rules! node_id_property_methods {
             ($id, $getter, get_node_id_property, Option<NodeId>, $setter, set_node_id_property, NodeId, $clearer)
         })*
         impl Node {
-            option_properties_debug_method! { debug_node_id_properties, $($getter)* }
+            option_properties_debug_method! { debug_node_id_properties, [$($getter,)*] }
         }
         impl NodeBuilder {
-            option_properties_debug_method! { debug_node_id_properties, $($getter)* }
+            option_properties_debug_method! { debug_node_id_properties, [$($getter,)*] }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


DataTriny:
I think I can use the macros even more. I'll update this later.

mwcampbell:
To be clear, my lack of response doesn't mean I have a problem with this PR; I just haven't taken time to review it yet.

DataTriny:
No worries, I just didn't want you to merge it as is and couldn't find how to set the PR as draft on the GitHub mobile app. Now I'm done.

mwcampbell:
In `option_properties_debug_method`, do we want to leave `value` wrapped and call `value.is_some`? Maybe it would be better to do something like this:

```
if let Some(value) = self.$getter() {
```

Then we can output the unwrapped value and have slightly less cluttered output.

mwcampbell:
The only other thing I don't like is the implementation of `debug_supported_actions`, and the accompanying manual implementation of `From<u8>` for `Action`. It requires all the variants to be repeated in the `From<u8>` implementation, and if a new variant is added, `debug_supported_action` has to be updated. All of that leaves room for mistakes. I understand why you did it this way; we don't want to add `enumn` as a required dependency of `accesskit`, since that would also bring in a dependency on the heavy `syn` crate. But I wonder if it would be possible to use a simple `macro_rules` macro to both define the enum and implement the `debug_supported_actions` function, with expanded code for testing all of the variants.

DataTriny:
My poor attempt at mimicking the enumflags2 crate feels a bit weird as we don't have a dedicated type to represent the mask. I think we can go without it for now.

Here's how the output of the example above looks like with the latest commit:

```rust
NodeBuilder { role: Button, actions: [Focus], is_busy: true, name: "Send", default_action_verb: Click, bounds: Rect { x0: 0.0, y0: 0.0, x1: 50.0, y1: 200.0 } }
Node { role: Button, actions: [Focus], is_busy: true, name: "Send", default_action_verb: Click, bounds: Rect { x0: 0.0, y0: 0.0, x1: 50.0, y1: 200.0 } }
```

Let me know what you think @mwcampbell.

mwcampbell:
Looks good to me.

PR #454, opened 2024-09-22
fix: Return a null object for AT-SPI application's parent

DataTriny:
AT-SPI's documentation clarified this since I first implemented the Unix adapter: the root/application node's parent is not the desktop object but a null reference, and the bus name component of an object reference is an empty string.


mwcampbell (approved):
Looks good to me.

PR #456, opened 2024-09-30
fix: update to ZBus v4

fogti:
This also requires bumps to
* atspi        v0.19 -> v0.22
* atspi-common v0.3  -> v0.6
* zbus         v3.14 -> v4.4
* zvariant     v3    -> v4.2

This is motivated by wanting to reduce the amount of dependencies on `syn` v1.

### DIFF
--- a/platforms/unix/src/atspi/object_address.rs
+++ b/platforms/unix/src/atspi/object_address.rs
@@ -3,10 +3,10 @@
 // the LICENSE-APACHE file) or the MIT license (found in
 // the LICENSE-MIT file), at your option.
 
-use atspi::Accessible;
+use atspi::ObjectRef;
 use serde::{Deserialize, Serialize};
 use zbus::{
-    names::UniqueName,
+    names::{OwnedUniqueName, UniqueName},
     zvariant::{ObjectPath, OwnedObjectPath, OwnedValue, Type, Value},
 };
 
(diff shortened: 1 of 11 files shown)


DataTriny:
Hello @fogti and thank you for working on this.

We first needed to increase the MSRV to 1.75, this is the main reason why I didn't do this before. #457 was just merged, can you please rebase?

platforms/unix/src/atspi/object_address.rs
```diff
@@ -3,39 +3,39 @@
...
-            bus_name: bus_name.to_string(),
+            bus_name: bus_name.to_owned().into(),
             path,
         }
     }
 
     pub(crate) fn null() -> Self {
         Self {
-            bus_name: String::new(),
+            bus_name: UniqueName::from_static_str("").unwrap().into(),
```
DataTriny, line 31: This will fail as D-Bus unique names can't be empty strings. Unless there is a reason to these changes that I don't get, please revert changes to this file.

fogti: ok, I was irritated that the `atspi` crate uses completely different values for the "null" object address, I can use their values instead.

fogti: specifically: https://docs.rs/atspi-common/0.6.0/src/atspi_common/object_ref.rs.html#23-32

DataTriny (approved):
The winit example still runs fine.

mwcampbell:
Looks good now, thanks.

@DataTriny If you agree, please merge when CI is done.

DataTriny:
Thank you again @fogti!

Issue #460, opened 2024-10-08
0.16.2 release broken?


trevyn:
0.16.1 worked fine, getting this in CI:

```
error[E0592]: duplicate definitions with name `n`
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/accesskit-0.16.2/src/lib.rs:270:38
    |
270 | #[cfg_attr(feature = "enumn", derive(enumn::N))]
    |                                      ^^^^^^^^ duplicate definitions for `n`
...
349 |     fn n(value: u8) -> Option<Self> {
    |     ------------------------------- other definition for `n`
    |
    = note: this error originates in the derive macro `enumn::N` (in Nightly builds, run with -Z macro-backtrace for more info)
error[E0034]: multiple applicable items in scope
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/accesskit-0.16.2/src/lib.rs:386:39
    |
386 |     while let Some(variant) = Action::n(i) {
    |                                       ^ multiple `n` found
    |
note: candidate #1 is defined in an impl for the type `Action`
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/accesskit-0.16.2/src/lib.rs:270:38
    |
270 | #[cfg_attr(feature = "enumn", derive(enumn::N))]
    |                                      ^^^^^^^^
note: candidate #2 is defined in an impl for the type `Action`
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/accesskit-0.16.2/src/lib.rs:349:5
    |
349 |     fn n(value: u8) -> Option<Self> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: this error originates in the derive macro `enumn::N` (in Nightly builds, run with -Z macro-backtrace for more info)
```

PR #463, opened 2024-10-08
fix: enhance workflows

bircni:
Improved the workflow so that #460 should not happen again


DataTriny:
Hello @bircni,

What led to #460 is that the CI doesn't check the accesskit crate with the enumn feature enabled. So, while this PR goes in the right direction, it will require more work to ensure such thing doesn't happen anymore.

PR #464, opened 2024-10-08
fix: Update some dependencies to remove duplicates

DataTriny:
We are able to remove all duplicates by updating the following crates:

- async-channel to 2.3.1
- enumn to 0.1.14
- schemars to 0.8.21
- thiserror to 1.0.64
-

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -185,21 +185,20 @@ version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "20cd0e2e25ea8e5f7e9df04578dc6cf5c83577fd09b1a46aaf5c85e1c33f2a7e"
 dependencies = [
- "event-listener 5.3.1",
- "event-listener-strategy 0.5.2",
+ "event-listener",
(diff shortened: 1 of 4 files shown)


fogti:
It shouldn't be necessary to force patch bumps via `Cargo.toml` tho, bumping only `async-channel` in `Cargo.toml` and checking that it doesn't affect this new updated `Cargo.lock` should suffice.

DataTriny:
@fogti It surely would work but updating Cargo.toml directly is more explicit. Do you have a particular reason for pushing back on this?

mwcampbell:
Here's why I also think it would be better to just update our Cargo.lock and not the crate manifests:

Updating the version numbers in the crate manifests implies that the minimum versions we require all downstream users to use for those crates has changed. But that's not the case. We're just doing a bit of cleanup on our own dependency graph as specified in our own Cargo.lock. Also, this approach would let us downgrade this PR from a fix to a chore, not triggering a release.

DataTriny:
Reverted all my changes then used `cargo update` with all the crates mentioned above. We end up with the same lock file.

PR #474, opened 2024-10-27
Add isEnabled to MacOS adapter

HolgerGottChristensen:
Added isAccessibilityEnabled to MacOS adapter.

This is based on the is_disabled attribute from the consumer crate.

If an accessibility node is explicitly set to disabled, it is marked as disabled on the MacOS accessibility node.
Otherwise it is marked as enabled.

When read out by VoiceOver, the item is read as "dimmed".

https://github.com/user-attachments/assets/89286e40-c532-465a-a39f-14294da3ac60

### DIFF
--- a/platforms/macos/src/node.rs
+++ b/platforms/macos/src/node.rs
@@ -524,7 +524,7 @@ declare_class!(
         }
 
         #[method(isAccessibilityEnabled)]
-        fn enabled(&self) -> bool {
+        fn is_enabled(&self) -> bool {
             self.resolve(|node| !node.is_disabled()).unwrap_or(false)
         }
 
--- a/platforms/macos/src/node.rs
+++ b/platforms/macos/src/node.rs
@@ -525,7 +525,7 @@ declare_class!(
 
         #[method(isAccessibilityEnabled)]
         fn enabled(&self) -> bool {
-            self.resolve(|node| !node.is_disabled()).unwrap_or(true)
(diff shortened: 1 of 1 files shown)


platforms/macos/src/node.rs
```diff
@@ -523,6 +523,11 @@ declare_class!(
                 .unwrap_or(false)
         }
 
+        #[method(isAccessibilityEnabled)]
+        fn enabled(&self) -> bool {
+            self.resolve(|node| !node.is_disabled()).unwrap_or(true)
```
DataTriny, line 528: If, for whatever reason, the node cannot be resolved, it should be marked as disabled. At least this is what Chromium does.

```suggestion
            self.resolve(|node| !node.is_disabled()).unwrap_or(false)
```

DataTriny (changes requested):
Thanks @HolgerGottChristensen for looking into this.

platforms/macos/src/node.rs
```diff
@@ -523,6 +523,11 @@ declare_class!(
                 .unwrap_or(false)
         }
 
+        #[method(isAccessibilityEnabled)]
+        fn enabled(&self) -> bool {
```
DataTriny, line 527: Also, `is_enabled` would be more appropriate. Please rename this method and update below.

DataTriny (changes requested):

DataTriny (approved):

PR #475, opened 2024-10-29
refactor!: Rename `name` to `label` and use `value` for label content

mwcampbell:
I think the new property name and the new use of `value` on nodes with `Role::Label` makes sense because:

* For non-label nodes, their label is now determined by either a direct string in the `label` property, or the `labelled_by` relation. The word choice is more consistent.
* On a label itself, it doesn't make sense to talk about the "label of a label", but "value of a label" does make sense.
* For labels that have full text support via `TextRun` nodes, we can now automatically get the label's value from its descendant text runs, rather than having to set the value directly; this follows automatically from the existing support for constructing the value from text in `accesskit_consumer::Node::value`.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -784,7 +784,7 @@ enum PropertyId {
     PopupFor,
 
     // String
-    Name,
+    Label,
     Description,
     Value,
     AccessKey,
@@ -1537,7 +1537,7 @@ node_id_property_methods! {
 }
 
 string_property_methods! {
-    (Name, name, set_name, clear_name),
+    (Label, label, set_label, clear_label),
     (Description, description, set_description, clear_description),
     (Value, value, set_value, clear_value),
     /// A single character, usually part of this node's name, that can be pressed,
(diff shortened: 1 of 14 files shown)


DataTriny:
I'm on board with these changes, but I expect the transition to the value property to be painful. Maybe you should split this PR so it has its dedicated entry in the changelog, until we properly document everything?

Also, you're effectively creating a new kind of nodes which are labeled by their value, and I wonder if you should explicitly name them by adding a new method in the consumer crate. I currently have no idea how we'd name this class of nodes, but I don't really like having this check on roles in multiple places.

mwcampbell:
I don't think I understand the benefit of splitting the PR.

mwcampbell:
I don't like the extra churn that splitting the PR would cause in the code itself, just for one extra line in the change log.

DataTriny:
Thanks for adding documentation. It doesn't mention that it also applies to `Role::Image` though. This actually make me question the purpose of this: if I hear "the value of an image", I won't think "text alternative". I'd say that the label property is more appropriate for this role. What was your reasoning for including the image role in this change?

DataTriny:
Maybe this is to allow a rich text alternative?

mwcampbell:
You're right about `Role::Image`. I fixed that, and also addressed your other suggestion.

DataTriny (approved):
Awesome, thanks! Please rebase and merge after that.

PR #476, opened 2024-10-29
refactor!: Rename `NodeBuilder` to `Node` and the old `Node` to `FrozenNode`

mwcampbell:
By doing this, passing the new `Node` (former `NodeBuilder`) in `TreeUpdate`, and having the consumer convert to the slightly smaller `FrozenNode`, we make the API more convenient for all users while not giving up the node size optimization. And it's probably slightly better for binary size to do the conversion to `FrozenNode` in just one place.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -1799,7 +1799,7 @@ vec_property_methods! {
 
 impl fmt::Debug for FrozenNode {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let mut fmt = f.debug_struct("Node");
+        let mut fmt = f.debug_struct("FrozenNode");
 
         fmt.field("role", &self.role());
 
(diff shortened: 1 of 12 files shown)


common/src/lib.rs
```diff
@@ -1779,19 +1785,19 @@ property_methods! {
...
 }
 
 vec_property_methods! {
     (CustomActions, CustomAction, custom_actions, get_custom_action_vec, set_custom_actions, set_custom_action_vec, push_custom_action, push_to_custom_action_vec, clear_custom_actions)
 }
 
-impl fmt::Debug for Node {
+impl fmt::Debug for FrozenNode {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let mut fmt = f.debug_struct("Node");
```
DataTriny, line 1802: Did you not change the name of the struct on purpose? I think we still want the debug representation to show if it's a frozen node or not, right?

DataTriny (approved):
I think this will even make our Python bindings less awkward.

PR #486, opened 2024-12-08
fix: Optimize removal of unreachable nodes

mwcampbell:
This eliminates an unnecessary step involving temporary storage. I originally wrote that code when I was still quite new to Rust and never revisited the assumption that `to_remove` was unnecessary.

I also renamed the variable that tracks unreachable nodes and the function that traverses them.


DataTriny (approved):
Looks good to me. Please rebase.

PR #487, opened 2024-12-08
chore: Update some dependencies to fix potential vulnerabilities

DataTriny:
- Updates tempfile (`cargo update tempfile`), which removes dependency on the instant crate, which fixes [RUSTSEC-2024-0384](https://rustsec.org/advisories/RUSTSEC-2024-0384).
- Updates num_enum (`cargo update num_enum`) which removes an older version of hashbrown from the dependency tree.
- Updates hashbrown (`cargo update hashbrown`) which fixes [RUSTSEC-2024-0402](https://rustsec.org/advisories/RUSTSEC-2024-0402).

Clippy fixes will come in a separate PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,7 +45,7 @@ name = "accesskit_consumer"
 version = "0.26.0"
 dependencies = [
  "accesskit",
- "hashbrown 0.15.0",
+ "hashbrown",
  "immutable-chunkmap",
 ]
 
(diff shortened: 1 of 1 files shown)


mwcampbell:
Since we're just updating Cargo.lock and not making any actual changes to any of the crates' Cargo.toml files, I think we can change this from a fix to a chore and eliminate the forced dependency bumps. The purpose of this is just to fix our own CI; we don't need to publish any new crate versions for this.

DataTriny:
You're right now that the bindings have moved. I'll revert these changes.

mwcampbell (approved):
There are several nice simplifications to our Cargo.lock in here.

PR #492, opened 2024-12-08
refactor!: Drop `Tree::app_name`

mwcampbell:
I now believe this property was a mistake.

On platforms where assistive technologies have conditional behavior based on the app name, they should get the app name from the platform, not the app itself. This way, correct behavior isn't dependent on the app or toolkit exposing the app name, and we don't give the app the opportunity to lie about its name.

Yes, AT-SPI requires the app to provide its own app name. I consider that a flaw in AT-SPI, which we should work around in the Unix adapter, rather than perpetuating it in our new cross-platform API. In Newton (the new free desktop accessibility stack I prototyped earlier this year), I want ATs to get the app name from the compositor, not the app itself, though I didn't implement that yet.

While I was here, I changed the `toolkit_name` and `toolkit_version` accessors in `accesskit_consumer` to return `Option<&str>` rather than cloning the `Option<String>`, since the latter is an unnecessary pessimization.

### DIFF
--- a/platforms/windows/src/node.rs
+++ b/platforms/windows/src/node.rs
@@ -645,9 +645,7 @@ impl IRawElementProviderSimple_Impl for PlatformNode_Impl {
                 }
                 match property_id {
                     UIA_FrameworkIdPropertyId => result = state.toolkit_name().into(),
-                    UIA_ProviderDescriptionPropertyId => {
-                        result = app_and_toolkit_description(state).into()
-                    }
+                    UIA_ProviderDescriptionPropertyId => result = toolkit_description(state).into(),
                     _ => (),
                 }
             }
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -2240,8 +2240,6 @@ impl JsonSchema for Properties {
 pub struct Tree {
     /// The identifier of the tree's root node.
     pub root: NodeId,
-    /// The name of the application this tree belongs to.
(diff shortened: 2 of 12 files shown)


platforms/windows/src/node.rs
```diff
@@ -645,9 +645,7 @@ impl IRawElementProviderSimple_Impl for PlatformNode_Impl {
                 }
                 match property_id {
                     UIA_FrameworkIdPropertyId => result = state.toolkit_name().into(),
-                    UIA_ProviderDescriptionPropertyId => {
-                        result = app_and_toolkit_description(state).into()
-                    }
+                    UIA_ProviderDescriptionPropertyId => result = toolkit_description(state).into(),
```
DataTriny, line 648: In fact, we should remove this property entirely. According to MSDN `UIA_ProviderDescription` is read-only and I can't see the value on Accessibility Insights.

mwcampbell: I can see the `ProviderDescription` value in NVDA using the NVDA+F1 command. That property does include the toolkit version, unlike `FrameworkId`. But I'm OK with removing it if you think that's better.

DataTriny: I am sure I've read on the documentation that this property wasn't meant to be set by providers but I can't find this back. No reason to drop it then.

DataTriny (approved):

PR #494, opened 2024-12-13
refactor!: Unbox the text selection property

mwcampbell:
`TextSelection` used to be much bigger when `NodeId` was 128-bit. But now we can unbox that property without increasing the total size of `PropertyValue`. So this PR does that. Also, I decided that the `rect` accessor should return a reference to the `Rect` struct rather than a copy. This makes the accessors for large structs more consistent.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -753,7 +753,7 @@ enum PropertyValue {
     VerticalOffset(VerticalOffset),
     Affine(Box<Affine>),
     Rect(Rect),
-    TextSelection(Box<TextSelection>),
+    TextSelection(TextSelection),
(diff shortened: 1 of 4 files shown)


mwcampbell:
I no longer think I want to do this. Instead, I'm thinking about moving the `rect` property to a direct field on `Node`. That would increase the size of `Node` by 32 bytes, but would decrease the size of `PropertyValue` by 8 bytes, from 40 to 32. That would allow property values to be a little more densely packed, with two property values fitting in a typical 64-byte cache line. And nearly every node has a bounding rectangle, so changing that property into a direct field would eliminate a slot in the `Vec` of dynamic property values for most nodes, meaning that the initial allocation for that `Vec` would go a little further.

PR #496, opened 2024-12-14
refactor!: Drop `FrozenNode`

mwcampbell:
If we decide to proceed in the direction started by #495, that is, minimizing temporary allocations when updating the tree at the cost of maintaining two copies of the tree state, then this PR is the next logical step. Dropping `FrozenNode` allows us to update nodes in place, including extending the `Vec` of properties, at the cost of 8 extra bytes per retained node (increasing the size from 120 to 128).

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -884,23 +884,6 @@ impl Default for PropertyIndices {
     }
 }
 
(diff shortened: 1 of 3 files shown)


DataTriny (approved):
Looks good to me but it needs to be rebased.

PR #505, opened 2025-01-16
[unix] Rewrite documentation for update_window_focus_state

drewcrawford:
Documenting the many misuses of this function that have been presenting as "orca doesn't seem to do anything".

### DIFF
--- a/platforms/unix/src/adapter.rs
+++ b/platforms/unix/src/adapter.rs
@@ -162,7 +162,48 @@ impl Adapter {
         }
     }
 
-    /// Update the tree state based on whether the window is focused.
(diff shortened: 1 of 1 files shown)


DataTriny:
Once again, sorry @drewcrawford for the delayed response. While there is valuable information in this pull request, I don't think much of it belong in the documentation of a method. You are right that what we have currently doesn't reflect the importance of calling this method. I think it would be more appropriate to explain the general call flow on the struct's documentation. I'd also prefer not directly mentioning Orca in our documentation as we strive to be generic.

I'd feel bad for asking changes from you after that much time has elapsed. I'm going to close this now. I really appreciate the effort you put into this though.

PR #508, opened 2025-01-26
Update objc2 related crates

DataTriny:


mwcampbell:
I wonder if it's too soon to do this. If we don't upgrade our objc2 dependencies in lock-step with the latest released version of winit, someone's cargo-deny is likely to complain.

DataTriny:
I agree that we should get this with the update to winit 0.31 probably.

hacknus:
Can this get priority? `winit` is already using `objc2` 0.6.1 but still depending on `accesskit_winit` which depends on 0.5.2

DataTriny:
The latest version of winit is [still using objc2 0.5](https://github.com/rust-windowing/winit/blob/v0.30.x/Cargo.toml). The plan is to merge this once we update to winit 0.31.

BTW winit does not depend on accesskit_winit. I assume you are depending on the winit master branch, in which case we can't help you at this time. Sorry.

PR #511, opened 2025-02-02
fix: Update winit to 0.30.8

### LINKED ISSUE #499: winit 0.30.6 broke AccessKit on Windows
To reproduce, in this repo, update winit to 0.30.6 with this command:

```
cargo update --precise 0.30.6 winit
```

Then build and run the `simple` example under `platforms/winit`. The focused window isn't announced. Updating to 0.30.5 using an appropriately modified version of the above command shows that 0.30.5 didn't have this problem.

I'm treating this issue as critical. I'm treating it as an AccessKit bug until we know for sure that it can only be fixed in winit.

DataTriny:
Fixes #499 

I just ran `cargo update --precise 0.30.8 winit`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -143,7 +143,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ef6978589202a00cd7e118380c448a08b6ed394c3a8df3a430d0898e3a42d046"
 dependencies = [
  "android-properties",
- "bitflags 2.4.1",
+ "bitflags 2.8.0",
  "cc",
  "cesu8",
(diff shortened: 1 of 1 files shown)


mwcampbell:
I noticed that this had the side effect of downgrading the windows-targets version that libloading depends on in our Cargo.lock. I guess this happened because libloading depends on a range of windows-targets versions. This looks safe to fix by manually editing Cargo.lock to revert that one change.

DataTriny:
Yeah, I had noticed this as well. Let me check what I can do.

mwcampbell (approved):

PR #512, opened 2025-02-02
fix: Update pyo3 to 0.23

DataTriny:
This will allow us to support Python 3.13 for our Python bindings.


DataTriny:
I'll go ahead and merge this. I don't think there are changes I could have done differently.

PR #513, opened 2025-02-05
fix: Derive `Debug` for adapters

drewcrawford:
Derive the Debug trait for the unix platform adapter.

Side note, I now have [code](https://github.com/drewcrawford/app_window) in the wings that relies on my PRs.  I need to explore alternatives to relying on my patches landing here.


DataTriny (approved):
Thank you very much.

PR #516, opened 2025-02-23
fix: Expose password input fields on Windows

DataTriny:
We already expose password/protected fields on other platforms, so this just bring the Windows adapter on par.


MarcoZehe (approved):
Looks good to me.

mwcampbell (approved):

PR #517, opened 2025-02-24
refactor!: Add event loop parameter to winit adapter constructors

mwcampbell:
I did this as part of the work on the Android adapter, but I think we should separate this out into its own PR so it has its own changelog entry and is explicitly pointed out as a breaking change.


DataTriny (approved):
This should definitely be its own entry in the Changelog.

PR #521, opened 2025-02-26
fix: Don't enable the `zbus/tokio` feature on Unix

### LINKED ISSUE #515: can't use tokio feature if a transient depends on accesskit with default-features
8f937ba makes default-features incompatible with the tokio feature
Talon depends on zbus with the tokio feature, so it needs to depend on accesskit with the tokio feature
Talon depends on egui
egui depends on accesskit with default-features (egui shouldn't need to decide about accesskit_unix's executor anyway)

Talon -> egui -> accesskit (default-features=true) enables the async-io accesskit feature
Talon -> accesskit (features=[tokio]) enables the tokio feature

accesskit now denies this:
```
error: Both "async-io" (default) and "tokio" features cannot be enabled at the same time.
```

this kind of defeats the purpose of the tokio feature in accesskit imo, as I shouldn't be expected to patch egui for default-features = false to use it

----

anyway, at this point I think nobody should use the zbus/tokio feature in a library crate, and we should instead use [this](https://github.com/dbus2/zbus/blob/main/zbus/README.md#compatibility-with-async-runtimes) approach for tokio support when the tokio feature is enabled in accesskit_unix

here's the commit where I did it for ksni: https://github.com/talonvoice/ksni/commit/15e31db06a203b061f57344a61f85cf45524650a

for now I'm turning off accesskit/tokio, and I wouldn't mind turning it back on with the zbus tokio feature removed like in the ksni change

DataTriny:
Fixes #515 

For simplicity, I chose to also drop the path to spawn a task using the unstable API. We'd have to enable the `tokio_unstable` flag ourself (previously done by zbus), just to name the task. Yes this would be useful in diagnozing a crash, but we don't test this enough for me to feel confident keeping this unstable code around.

Tested by running the winit example with both runtimes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1873,14 +1873,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e2b070231665d27ad9ec9b8df639893f46727666c6767db40317fbe920a5d998"
 dependencies = [
  "backtrace",
- "bytes",
  "libc",
  "mio",
  "pin-project-lite",
- "signal-hook-registry",
  "socket2",
  "tokio-macros",
- "tracing",
  "windows-sys 0.52.0",
 ]
 
@@ -2687,7 +2684,6 @@ dependencies = [
  "serde",
  "serde_repr",
  "static_assertions",
- "tokio",
  "tracing",
  "uds_windows",
  "windows-sys 0.59.0",
--- a/platforms/unix/Cargo.toml
+++ b/platforms/unix/Cargo.toml
@@ -13,16 +13,16 @@ rust-version.workspace = true
 
 [features]
 default = ["async-io"]
-async-io = ["dep:async-channel", "dep:async-executor", "dep:async-task", "dep:futures-util", "atspi/async-std", "zbus/async-io"]
-tokio = ["dep:tokio", "dep:tokio-stream", "atspi/tokio", "zbus/tokio"]
+async-io = ["dep:async-channel", "dep:async-executor", "dep:async-task", "dep:futures-util"]
+tokio = ["dep:tokio", "dep:tokio-stream"]
 
 [dependencies]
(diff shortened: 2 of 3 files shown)


mwcampbell:
@lunixbochs Can you easily test this solution and make sure it does what you need? Thanks.

mwcampbell:
Given that `accesskit_unix` doesn't offer an async API at all, but always does its async work on another thread, and that it's now always using `zbus`'s `async-io` feature, do we even need to offer `async-io` and `tokio` feature, or should we just always use `async-io`? @lunixbochs What do you think?

DataTriny:
Even when the `zbus/async-io` feature is enabled, we still spin a tokio executor if the `tokio` feature is enabled since we don't use zbus's blocking API. It's just that we can't build zbus without picking an async runtime. So I think it's just that tokio users will depend on some async-std dependencies and will have to explicitly depend on accesskit_unix to disable default features.

But I agree that dropping support for tokio entirely would make things simpler though.

mwcampbell:
@lunixbochs Can you please take a look at this? Thanks.

lunixbochs:
confirmed to coexist happily with my ksni fork under all configurations

ksni (asyncio or tokio)
accesskit_unix (asyncio or tokio)

if I turn on the zbus/tokio feature, accesskit_unix must have the tokio feature enabled or zbus will panic internally, but this is strictly better than the previous behavior

the only thing remaining is that enabling the tokio feature in accesskit_unix is not possible while the default async-io feature is enabled. I think it would be slightly easier for folks who end up with an accesskit_unix dependency in their tree to work around the zbus/tokio feature if you could turn on both async-io and tokio at the same time and it would make sure to use a tokio executor, but that's not a dealbreaker.

mwcampbell:
In that case, I propose we merge and release.

PR #527, opened 2025-03-09
fix: Replace paste by the compose-idents crate

### LINKED ISSUE #526: Paste dependency no longer maintained
```
error[unmaintained]: paste - no longer maintained
    ┌─ /home/runner/work/bevy/bevy/Cargo.lock:369:1
    │
369 │ paste 1.0.15 registry+https://github.com/rust-lang/crates.io-index
    │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
    │
    ├ ID: RUSTSEC-2024-0436
    ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0436
    ├ The creator of the crate `paste` has stated in the [`README.md`](https://github.com/dtolnay/paste/blob/master/README.md) 
      that this project is not longer maintained as well as archived the repository
    ├ Announcement: https://github.com/dtolnay/paste
    ├ Solution: No safe upgrade is available!
    ├ paste v1.0.15
      ├── accesskit_windows v0.24.1
      │   └── accesskit_winit v0.23.1
```

Used in https://github.com/AccessKit/accesskit/blob/5328b23bb20a05404cf078cf20295abc194e36f2/platforms/windows/src/node.rs#L18

DataTriny:
Fixes #526 

The Windows adapter relies pretty heavily on this dependency and removing it would make maintenance and readability harder. Vendoring the macro is not an option either because it calls a procedural macro internally.

There are a few alternative crates but compose-idents seem to be the only one still maintained. It is MIT licensed.

Because you have to declare which identifiants you are going to use upfront, using its macro inside match arms is not possible. I have replaced these call sites by `if` statements. I have used cargo-show-asm to see the impact of this change on the generated assembly. I think there is no difference, but I'd appreciate a second check.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,7 +37,6 @@ dependencies = [
  "jni",
  "log",
  "once_cell",
- "paste",
 ]
 
 [[package]]
@@ -97,9 +96,9 @@ version = "0.25.0"
 dependencies = [
  "accesskit",
  "accesskit_consumer",
+ "compose-idents",
  "hashbrown",
  "once_cell",
- "paste",
  "scopeguard",
  "static_assertions",
  "windows",
@@ -299,7 +298,7 @@ checksum = "3b43422f69d8ff38f95f1b2bb76517c91589a924d1559a0e935d7c8ce0274c11"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.79",
 ]
 
 [[package]]
@@ -334,7 +333,7 @@ checksum = "721cae7de5c34fbb2acd27e21e6d2cf7b886dce0c27388d46c4e6c47ea4318dd"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.79",
 ]
 
 [[package]]
@@ -530,6 +529,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "compose-idents"
+version = "0.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2ac1c5875282ad51cd73e98b73d1827dbe46489da16184c14842b9652c0c6d46"
+dependencies = [
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
 name = "concurrent-queue"
 version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


DataTriny:
[pastey](https://github.com/AS1100K/pastey) was just published. It's apparently a drop-in replacement for paste. Like compose-idents it is very young so it is hard to tell which one will stick around at this time. The only thing that bothers me with compose-idents is that it relies on an older version of syn.

mwcampbell:
As you pointed out, pastey and compose-idents are both very young. They both have very few total downloads and no public dependents. And compose-idents depends on syn 1.x, causing a regression in efforts to eliminate multiple versions of dependencies.

Given all of that, I think the least bad option is to not depend on either crate, and explicitly write out all the variations of the identifiers in our macro invocations. It's tedious, but when given the choice to do a little extra work to eliminate problems, even if they're only perceived and not real, we should take it. We already did this in the accesskit crate itself, writing out identifiers rather than using paste, to get to zero mandatory dependencies for that crate.

AndreiPashkin:
@DataTriny, could you please clarify this use case in more detail?

> Because you have to declare which identifiants you are going to use upfront, using its macro inside match arms is not possible.

DataTriny:
@AndreiPashkin I was referring to this kind of patterns where we were building match arms for each repetition of a fragment:

```rust
match property_id {
    $(paste! { [< UIA_ $base_id PropertyId>] } => {
        self.$m().into()
    })*
    _ => Variant::empty()
}
```

The `compose_idents` macro would not be allowed where the `paste` macro was. In our specific case I was able to replace these by consecutive `if` statements at no cost, but if your `match` expression was optimized into a jump table by rustc then you would notice an impact on performance.

AndreiPashkin:
> The compose_idents macro would not be allowed where the paste macro was.

I think it would:
```rust
match property_id {
    $(compose_idents!(property_id_ident = [UIA_, $base_id, PropertyId]; { property_id_ident }) => {
        self.$m().into()
    })*
    _ => Variant::empty()
}
```

@DataTriny

DataTriny:
Actually it does, my bad. I don't exactly recall what I tried because I couldn't get it to compile last time.

Anyway, we decided that we could live with some extra identifiants in a few places. Thanks for offering compose-idents to the community though!

AndreiPashkin:
@DataTriny, no problem. One of my goals was to create a macro that is IDE-friendly and doesn't introduce a non-standard syntax (within the code block). I'll make it even cleaner in future versions.

PR #529, opened 2025-03-16
fix!: Panic if the window is visible when the adapter is created, for adapters where this is a problem

mwcampbell:
I've concluded that we have to enforce this rule in the most obnoxious way possible.

### DIFF
--- a/platforms/windows/src/subclass.rs
+++ b/platforms/windows/src/subclass.rs
@@ -149,11 +149,19 @@ impl SubclassingAdapter {
     /// This must be called on the thread that owns the window. The activation
     /// handler will always be called on that thread. The action handler
     /// may or may not be called on that thread.
+    ///
+    /// # Panics
+    ///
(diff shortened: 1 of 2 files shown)


DataTriny:
Although I was the one to initially suggest doing this, I now think the most correct and idiomatic way to address this would be to mark winit Adapter constructors as unsafe. We ask the caller to make sure the window is invisible before creating the adapter, bad things will happen if this condition is not met, and we don't take the responsibility of putting the system in a safe state ourself. I get that unsafe functions are annoying, but I think marking them unsafe and adding a safety paragraph to their documentation is the accepted way of dealing with cases like that.

DataTriny (changes requested):
I get that marking constructors as unsafe might be too controvertial as a change. Please add a panic chapter to the constructors instead.

DataTriny (approved):

Issue #349, opened 2024-02-01
Use `cargo-deny` to prevent duplicate dependencies


emilk:
By running [cargo-deny](https://github.com/EmbarkStudios/cargo-deny) on CI for all supported platforms, we can ensure users of accesskit won't have duplicated dependencies. Reducing duplicated dependencies reduces compile times and binary bloat.

Here's how it is done in `winit`, for instance:

https://github.com/rust-windowing/winit/blob/4d4d6e5052877d9578bac73f52eaf323e9089998/.github/workflows/ci.yml#L180-L203

and their `deny.toml`:

https://github.com/rust-windowing/winit/blob/master/deny.toml

emilk:
Oh sorry, I thought I checked first 🤦 

Yeah I got some duplicated deps via `accesskit-winit` and `accesskit-unix`… I'll see if I can reproduce

mwcampbell:
We haven't actually published a release since adding cargo deny.

emilk:
Ah, yes, the `deny.toml` does not prevent duplicated dependencies. What is needed is:

```diff
--- a/deny.toml
+++ b/deny.toml
@@ -34,7 +34,7 @@ exceptions = [
 ]
 
 [bans]
-multiple-versions = "warn"
+multiple-versions = "deny"
 wildcards = "allow"
 highlight = "all"
 workspace-default-features = "allow"
```

and then duplicates popup fast when checking different platforms with

```
cargo deny --all-features --log-level error --target aarch64-apple-darwin check
cargo deny --all-features --log-level error --target aarch64-linux-android check
cargo deny --all-features --log-level error --target i686-pc-windows-gnu check
cargo deny --all-features --log-level error --target i686-pc-windows-msvc check
cargo deny --all-features --log-level error --target i686-unknown-linux-gnu check
cargo deny --all-features --log-level error --target wasm32-unknown-unknown check
cargo deny --all-features --log-level error --target x86_64-apple-darwin check
cargo deny --all-features --log-level error --target x86_64-pc-windows-gnu check
cargo deny --all-features --log-level error --target x86_64-pc-windows-msvc check
cargo deny --all-features --log-level error --target x86_64-unknown-linux-gnu check
cargo deny --all-features --log-level error --target x86_64-unknown-linux-musl check
cargo deny --all-features --log-level error --target x86_64-unknown-redox check
```

Some are small things like `raw-window-handle` that should just be ignored, but there are also a lot of large crates that are duplicated.

[...]

PR #532, opened 2025-03-23
chore: Update dependencies to remove some duplicates

DataTriny:
Part of #349 

- `cargo update core-graphics-types`
- `cargo update mio`
- `cargo update polling`
- `cargo update tempfile`
- `cargo update toml_edit`


mwcampbell:
Is it necessary to downgrade libloading's dependency on windows-targets from 0.52 to 0.48, or can we try reverting that?

DataTriny:
It's apparently not. I'm not sure why the resolver chose to do that.

mwcampbell (approved):

Issue #444, opened 2024-08-14
Should accesskit use `Box<str>`?


PoignardAzur:
A lot of accesskit APIs end up using `Box<str>` as their string type, which is not very idiomatic for Rust.

As far as I can tell, the reason is to save space in the `PropertyValue` enum. But that enum also has a few variants that are as large or larger than a String, meaning storing a `Box<str>` doesn't actually save space:

- `NodeIdVec` has a Vec.
- The `Rect` variant holds 4 `f64`
- `CustomActionVec` has a Vec.

Should these variants be boxed? As far as I can tell every other type is two pointers long at most.

drewcrawford:
For what it's worth, I was momentarily confused by this today.  `String` feels more idiomatic for what I was doing, but I don't know the history.

PR #540, opened 2025-03-26
perf: Replace `Box<str>` with `Cow<'static, str>`

CrazyboyQCD:
Related with https://github.com/AccessKit/accesskit/issues/444.
`Box<str>` conversion requires one more possible reallocation in all variants of `String` except itself, we should use `Cow<'static, str>` to avoid that.
I'm not sure if this is a breaking change.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -12,7 +12,7 @@
 
 extern crate alloc;
 
-use alloc::{boxed::Box, string::String, vec::Vec};
+use alloc::{borrow::Cow, boxed::Box, string::String, vec::Vec};
 use core::fmt;
 #[cfg(feature = "pyo3")]
 use pyo3::pyclass;
@@ -656,7 +656,7 @@ impl From<NodeId> for NodeIdContent {
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 pub struct CustomAction {
     pub id: i32,
-    pub description: Box<str>,
+    pub description: Cow<'static, str>,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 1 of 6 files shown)


DataTriny:
Hello @CrazyboyQCD,

The reason why we keep `Box<str>` around is size. `Cow<'static, str>` is bigger by 8 bits. Furthermore, we have a big refactoring in the works that will affect the string getters and setters.

Your opinion @mwcampbell?

CrazyboyQCD:
Something like this should resolve the size issue:
```rust
#[repr(transparent)]
struct ReferentStr(str);

#[derive(Clone)]
#[repr(transparent)]
struct OwnedBoxedStr(Box<str>);

impl Borrow<ReferentStr> for OwnedBoxedStr {
    fn borrow<'s>(&'s self) -> &'s ReferentStr {
        // Use unsafe code to change type of referent.
        // Safety: `ReferentStr` is a `repr(transparent)` wrapper around `str`.
        let ptr = &*self.0 as *const str as *const ReferentStr;
        unsafe { &*ptr }
    }
}

impl ToOwned for ReferentStr {
    type Owned = OwnedBoxedStr;
    fn to_owned(&self) -> OwnedBoxedStr {
        OwnedBoxedStr(Box::from(&self.0))
    }
}

type CowBoxedStr = Cow<'static, ReferentStr>;
```

CrazyboyQCD:
@DataTriny 
Only size of `CustomAction` is increased by this, the other two `enum` doesn't get affected, if that's what is worried we could just restore `CustomAction`.

DataTriny:
Coming back to this.

The only time where this change actually improves the situation is when dealing with static `str`. While this is frequent in example programs, this rarely happens in the wild.

We are probably going to move to an API that copies string inputs systematically in the next version or so. Therefore I am going to close this.

Thank you for your contribution nonetheless @CrazyboyQCD.

PR #541, opened 2025-03-30
fix: Update windows to 0.61

DataTriny:
Recent changes in the `VARIANT` type make it easier to implement the `UIA_ControllerForPropertyId` property, which I need for dropdown.


DataTriny:
The only thing here I am unsure about is whether we really need to clone platform nodes to cast to a different interface. I might have missed something but I didn't see how to do it without cloning.

DataTriny:
Thanks for the fixes, they look good to me.

PR #544, opened 2025-04-06
fix: Update pyo3 to 0.24

DataTriny:
Addresses [RUSTSEC-2025-0020](https://rustsec.org/advisories/RUSTSEC-2025-0020)

Note that I didn't directly update to 0.24 in #512 because it brings more breaking changes, making the process of updating our Python bindings even more painful.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1443,9 +1443,9 @@ dependencies = [
 
 [[package]]
 name = "pyo3"
-version = "0.23.4"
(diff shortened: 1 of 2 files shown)


DataTriny:
The common crate compiles with the pyo3 feature and Cargo.lock only contain changes to pyo3-related crates and one of their dependencies.

I'm going to merge this without a review to fix CI.

PR #546, opened 2025-04-09
fix: Update tokio and tokio-stream dependencies

DataTriny:
Addresses [RUSTSEC-2025-0023](https://rustsec.org/advisories/RUSTSEC-2025-0023)

- `cargo update tokio-stream`
- `cargo update tokio`

For some reason, Cargo once again downgraded libloading's dependency on windows-targets so I had to revert this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1871,9 +1871,9 @@ dependencies = [
 
 [[package]]
 name = "tokio"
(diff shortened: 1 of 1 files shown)


DataTriny:
I am going to merge this now to fix CI.

PR #550, opened 2025-04-14
refactor!: Drop redundant `HasPopup::True`

DataTriny:
`HasPopup::True` is effectively an alias for `HasPopup::Menu`. Since it is less meaningful and redundant, I propose to drop this variant.


mwcampbell:
This might be a good PR in which to add dummy changes to Cargo.toml in all the crates that don't yet have semver-breaking version number bumps in the current release PR.

mwcampbell (approved):

PR #552, opened 2025-04-15
fix: Return text content from multiline inputs

mwcampbell:
`accesskit_consumer::Node::value` doesn't traverse a multiline input's descendants because we don't want to have the potential performance problem of traversing a large document every time a single character is typed. But I'm not aware of any other option on Android, so I changed the `text` method on `NodeWrapper` in the Android adapter.

### DIFF
--- a/platforms/android/src/node.rs
+++ b/platforms/android/src/node.rs
@@ -67,7 +67,11 @@ impl NodeWrapper<'_> {
     }
 
     pub(crate) fn text(&self) -> Option<String> {
-        self.0.value()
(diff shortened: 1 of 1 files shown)


DataTriny:
Is there an easy way to test this?

DataTriny (approved):
Looks good codewise.

mwcampbell:
There isn't yet an easy way to test this. There will be soon, when I've made a little more progress on the demo app for my [android-view](https://github.com/mwcampbell/android-view) project.

PR #553, opened 2025-04-16
fix: Fix Android adapter after dropping `FrozenNode`

mwcampbell:
It looks like the Android adapter isn't actually being built in CI. Let's fix that in a separate PR.


DataTriny (approved):
I did not try to compile your previous PR. This should have been caught at that time.

PR #555, opened 2025-04-18
refactor!: Use the queued-events pattern in the Android adapter

mwcampbell:
As with the Windows and macOS adapters, this lets us raise events while not holding any lock that may apply to the AccessKit adapter, to avoid deadlocks due to reentrancy. This is important for my [android-view](https://github.com/mwcampbell/android-view) project, which is already using this branch. Also, when a click action is performed, the corresponding event is now fired on the native side, so the Java code can be a little simpler.


DataTriny (approved):

DataTriny (changes requested):
Actually, the documentation for `update_if_active` should be modified to mention the return value, like we do in the other adapters.

DataTriny (approved):

PR #558, opened 2025-04-28
refactor!: Simplify the core Android adapter API

mwcampbell:
Now even the `Adapter` API prioritizes a simple interface and ease of integration over ease of internal implementation. And I believe it does so without meaningfully compromising performance. Specifically, this lower-level adapter still avoids the overhead of unnecessarily sending things across threads if the UI toolkit is already running on the Android UI thread.

The two adapters can be tested with the demo applications in these two repositories:

* `Adapter`: [android-view](https://github.com/mwcampbell/android-view); for the first time, this repo demonstrates a direct integration using `Adapter` that I'm happy with
* `InjectingAdapter`: [xilem-to-do-mvc-android](https://github.com/mwcampbell/xilem-to-do-mvc-android), an updated version of my test/demo app using GameActivity, winit, and xilem that was the only real test case for the original version of the Android adapter


platforms/android/src/adapter.rs
```diff
@@ -367,84 +237,111 @@ impl Adapter {
         }
     }
 
-    /// Populate an `AccessibilityNodeInfo` with information about the
-    /// AccessKit node identified by the given virtual view ID.
+    /// Create an `AccessibilityNodeInfo` for the AccessKit node
+    /// corresponding to the given virtual view ID. Returns null if
+    /// there is no such node.
     ///
     /// The `host` paramter is the Android view for this adapter.
```
DataTriny, line 244:
```suggestion
    /// The `host` parameter is the Android view for this adapter.
```

platforms/android/src/adapter.rs
```diff
@@ -367,84 +237,111 @@ impl Adapter {
...
-        let node = tree_state.focus_in_tree();
-        self.node_id_map.get_or_create_java_id(&node)
+        node_info
     }
 
-    pub fn virtual_view_at_point<H: ActivationHandler + ?Sized>(
+    /// Create an `AccessibilityNodeInfo` for the AccessKit node
+    /// with the given focus type. Returns null if there is no such node.
+    ///
+    /// The `host` paramter is the Android view for this adapter.
```
DataTriny, line 317:
```suggestion
    /// The `host` parameter is the Android view for this adapter.
```

platforms/android/src/adapter.rs
```diff
@@ -750,4 +636,114 @@ impl Adapter {
...
+        let point = Point::new(x.into(), y.into());
+        let point = root.transform().inverse() * point;
+        let node = root.node_at_point(point, &filter)?;
+        Some(self.node_id_map.get_or_create_java_id(&node))
+    }
+
+    /// Handle the provided hover event.
+    ///
+    /// The `action`, `x`, and `y` parameters must be retrieved from
+    /// the corresopnding properties on an Android motion event. These
```
DataTriny, line 698:
```suggestion
    /// the corresponding properties on an Android motion event. These
```

platforms/android/src/inject.rs
```diff
@@ -126,218 +95,200 @@ fn inner_adapter_from_handle(handle: jlong) -> Option<Arc<Mutex<InnerInjectingAd
...
+    drop(inner_adapter);
+    events.raise(&mut env, &host);
+    JNI_TRUE
 }
 
-fn delegate_class(env: &mut JNIEnv) -> Result<&'static JClass<'static>> {
+fn delegate_class(env: &mut JNIEnv) -> &'static JClass<'static> {
     static CLASS: OnceCell<GlobalRef> = OnceCell::new();
-    let global = CLASS.get_or_try_init(|| {
+    let global = CLASS.get_or_init(|| {
```
DataTriny, line 221: You could use a `LazyCell` here right? Or maybe even better, use `std::sync::OnceLock` until our MSRV permit us to use `LazyLock`, which would allow us to drop our dependency on `once_cell`.

DataTriny (changes requested):

DataTriny (approved):

PR #563, opened 2025-05-16
Add object safe APIs to the adapters

mikayla-maki:
Hello!

In GPUI, we use an object safe platform abstraction, which is incompatible with the existing `update_if_active` APIs. This PR fixes this by introducing 2 new methods that allow us to re-implement `update_if_active` at a higher level:

- `active(&self) -> bool`
- `pub fn update(&mut self, update: TreeUpdate) -> Option<QueuedEvents>`

### DIFF
--- a/platforms/macos/src/adapter.rs
+++ b/platforms/macos/src/adapter.rs
@@ -103,6 +103,25 @@ impl Adapter {
         Self { state }
     }
 
+    /// Check if the adapter is currently active.
+    pub fn active(&self) -> bool {
+        match self.state {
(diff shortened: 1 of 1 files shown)


DataTriny:
Hi @mikayla-maki,

This does not feel right to me. Could you please show how you plan to use this?

Have you looked at the accesskit_winit crate (in `platforms/winit`), to see a similar implementation of what you are likely trying to build?

Thanks.

mikayla-maki:
Hi @DataTriny, thanks for letting me know, I will close this for now and get back to you soon!

PR #568, opened 2025-06-05
fix: Raise events when adding or removing nodes due to filtering on Unix

DataTriny:
We already create and destroy nodes in the DBus tree, but we also need to send appropriate events or otherwise ATs have no way of figuring out the changes.


mwcampbell:
Did you mean to mark this as a draft, or is it ready for review?

PR #573, opened 2025-06-13
refactor!: Refactor actions for scrolling by discrete units

mwcampbell:
This drops the `ScrollBackward` and `ScrollForward` actions, leaving only the actions for left, right, up, and down. It also adds a `ScrollUnit` enum for specifying whether to scroll by an item/line or a full page.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -2603,7 +2603,11 @@ pub struct TreeUpdate {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "schemars", derive(JsonSchema))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
-#[repr(C)]
+#[cfg_attr(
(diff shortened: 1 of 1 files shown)


common/src/lib.rs
```diff
@@ -2600,6 +2597,22 @@ pub struct TreeUpdate {
     pub focus: NodeId,
 }
 
+/// The amount by which to scroll in the direction specified by one of the
+/// `Scroll` actions.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "schemars", derive(JsonSchema))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[repr(C)]
```
DataTriny, line 2606: Please also add the `pyclass` attribute to this enum.

DataTriny (changes requested):
Just one comment regarding Python support.

PR #574, opened 2025-06-13
feat: Consumer support for scrolling and clipping children

mwcampbell:
This exposes the scrolling and `clips_children` on the consumer `Node` struct. It also adds a new rule to the common filter to exclude clipped children, except for one clipped child at either end of a list (so that child can be scrolled into view).

### DIFF
--- a/consumer/src/filters.rs
+++ b/consumer/src/filters.rs
@@ -14,16 +14,25 @@ pub enum FilterResult {
     ExcludeSubtree,
 }
 
-fn filter_for_sibling_clip_check(node: &Node) -> FilterResult {
+fn common_filter_base(node: &Node) -> Option<FilterResult> {
     if node.is_focused() {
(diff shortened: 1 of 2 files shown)


consumer/src/filters.rs
```diff
@@ -223,4 +272,133 @@ mod tests {
             common_filter(&tree.state().node_by_id(NodeId(1)).unwrap())
         );
     }
+
+    #[test]
+    fn clipped_children() {
```
DataTriny, line 277: I feel like this test tests too many things at once, making it harder to understand if it fail in the future. Could you please try to split it so that each function contain less assertions?

DataTriny (changes requested):
Just one comment on the added test. I know we have some large tests like this, as does Chromium, but I think we should avoid them as much as possible.

PR #586, opened 2025-06-23
feat: Scrolling on Android

mwcampbell:
Specifically, this implements the actions for scroling backward and forward by one line/item, and sets the right class name for the `ScrollView` role, so TalkBack will request those actions when flicking left and right through a list. This can be tested by building and running the Masonry demo in the [test-virtual-scroll branch of android-view](https://github.com/rust-mobile/android-view/tree/test-virtual-scroll).

### DIFF
--- a/platforms/android/src/adapter.rs
+++ b/platforms/android/src/adapter.rs
@@ -10,7 +10,7 @@
 
 use accesskit::{
     Action, ActionData, ActionHandler, ActionRequest, ActivationHandler, Node as NodeData, NodeId,
-    Point, Role, TextSelection, Tree as TreeData, TreeUpdate,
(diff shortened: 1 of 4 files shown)


DataTriny (changes requested):
Only one minor comment so far, but I haven't tried running an example yet.

DataTriny:
Sadly I cannot run any android-view app on the device I have. It runs Android 11, which mean the call to `ViewConfiguration.getMultiPressTimeout` results in a crash. I haven't found a way to check for SDK version from Rust so I'm not sure if it's possible to conditionally call this at the moment.

mwcampbell:
I suggest that for now, you just modify your copy of android-view to replace this call with a dummy value. It won't matter for testing the Masonry demo on the test-virtual-scroll branch with TalkBack, since that demo doesn't respond to touch events anyway.

PR #589, opened 2025-06-26
fix!: Force a semver-breaking release

mwcampbell:
I forgot to do this workaround when making a semver-breaking change to the `accesskit` crate.


DataTriny (approved):
Please rebase and then merge. Thanks.

PR #594, opened 2025-07-09
feat!: Implement refactored `ScrollIntoView` action across desktop platforms

mwcampbell:
The old `ScrollTarget` action data wasn't a good fit for what any platform actually supported. So I replaced it with a new `ScrollHint` which is a copy of Chromium's `ScrollType` enum, which in turn is a copy of AT-SPI's `ScrollType`.

The Windows implementation is known to work with Narrator, and the macOS implementation is known to work with VoiceOver. I don't know if there's any way to get Orca to call the `ScrollTo` method outside of web content. I know that NVDA doesn't call `ScrollIntoView` outside of web content. All of this can be tested using my [virtual-scroll-a11y branch of xilem](https://github.com/linebender/xilem/tree/virtual-scroll-a11y). That's actually still using the scrolling-wip branch of AccessKit, but it's the same code. I'll delete that branch after this PR is merged and the Xilem branch is rebased.

### DIFF
--- a/platforms/macos/src/node.rs
+++ b/platforms/macos/src/node.rs
@@ -30,6 +30,8 @@ use std::rc::{Rc, Weak};
 
 use crate::{context::Context, filters::filter, util::*};
 
+const SCROLL_TO_VISIBLE_ACTION: &str = "AXScrollToVisible";
+
 fn ns_role(node: &Node) -> &'static NSAccessibilityRole {
     let role = node.role();
     // TODO: Handle special cases.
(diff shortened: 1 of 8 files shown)


platforms/macos/src/node.rs
```diff
@@ -961,6 +963,37 @@ declare_class!(
             .flatten()
         }
 
+        // We discovered through epxerimentation that when mixing the newer
```
DataTriny, line 966:
```suggestion
        // We discovered through experimentation that when mixing the newer
```

DataTriny (changes requested):
Just found one typo. Feel free to merge once it's fixed.

Issue #595, opened 2025-07-14
Crash on Windows with gtk4


lazka:
Downstream report: https://gitlab.gnome.org/GNOME/gtk/-/issues/7653

Changing the visibility of a window leads to a crash in accesskit.

Using accesskit-c 0.16 (I patched gtk4 to check if a newer version changes things, but it didn't). If there is anything I can try, let me know. Or point me to the right place if this isn't it.

Stack trace:

```cpp
[endless loop until crash]
#2223 0x00007ffc21b7cb0d in accesskit_rect_is_empty () from C:\msys64\clang64\bin\libaccesskit-c-0.16.dll
#2224 0x00007ffc8f3681d6 in USER32!CallWindowProcW () from C:\WINDOWS\System32\user32.dll
#2225 0x00007ffc8f367c9c in USER32!CallWindowProcW () from C:\WINDOWS\System32\user32.dll
#2226 0x00007ffc21b7cb0d in accesskit_rect_is_empty () from C:\msys64\clang64\bin\libaccesskit-c-0.16.dll
#2227 0x00007ffc8f3681d6 in USER32!CallWindowProcW () from C:\WINDOWS\System32\user32.dll
#2228 0x00007ffc8f367c9c in USER32!CallWindowProcW () from C:\WINDOWS\System32\user32.dll
#2229 0x00007ffc21b7cb0d in accesskit_rect_is_empty () from C:\msys64\clang64\bin\libaccesskit-c-0.16.dll
#2230 0x00007ffc8f3681d6 in USER32!CallWindowProcW () from C:\WINDOWS\System32\user32.dll
#2231 0x00007ffc8f367a5c in USER32!SendMessageW () from C:\WINDOWS\System32\user32.dll
#2232 0x00007ffc8f39edb6 in USER32!FindWindowW () from C:\WINDOWS\System32\user32.dll
#2233 0x00007ffc8fbe5df4 in ntdll!KiUserCallbackDispatcher () from C:\WINDOWS\SYSTEM32\ntdll.dll
#2234 0x00007ffc8d6a16c4 in win32u!NtUserSetWindowPos () from C:\WINDOWS\System32\win32u.dll
#2235 0x00007ffc0b2816ba in gdk_win32_surface_resize (surface=0x83f69e0, width=358, height=268)
    at ../gdk/win32/gdksurface-win32.c:867
[...]

PR #596, opened 2025-07-16
fix: Panic if the subclassing adapter is double-instantiated

mwcampbell:
This is in response to #595.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -96,6 +96,7 @@ dependencies = [
  "accesskit_consumer",
  "hashbrown",
  "once_cell",
+ "parking_lot",
  "scopeguard",
  "static_assertions",
  "windows",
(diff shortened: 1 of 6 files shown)


DataTriny (approved):
Exactly what I had in mind, thank you. I appreciate the presence of an integration test.

Do you think we should do the same for the macOS subclassing adapter? I imagine that a double instantiation could also wreak havoc over there.

mwcampbell:
I implemented the same check on macOS, but I don't know how to write a test for it, because an NSWindow or NSView has to be instantiated on the process's main thread, and cargo tests apparently don't run on the main thread even if there's only one test in the crate.

DataTriny:
Thanks. Well this is unfortunate and we will have to find a solution for tests on macOS anyway. But I am fine not having a test for this now.

PR #616, opened 2025-09-17
update objc2 crates

httnn:
this pull request updates the objc2 ecosystem dependencies to their latest versions. the new versions include some useful features like being able to downcast `NSObject`s more easily.

most changes are due to renamed structs (`Id` -> `Retained`, `WeakId` -> `Weak`, `NSArray::from_vec` -> `NSArray::from_retained_slice`) and a slightly more streamlined way of defining new classes.

an open question: do we need to try to handle errors when working with `CString`s or is unwrapping fine? i guess since `CString`s are used only with class names unwrapping could be acceptable.

i have tested these changes to work but might be nice if someone else could confirm this as well.


httnn:
oops i missed #508! there are no conflicts with main on here but feel free to close anyway. was luckily a very quick thing to do 😄

DataTriny:
Sorry @httnn, it's always better to check for existing PRs before working on something. With that being said, I'll keep this one instead since it is more up to date. Thank you for your work!

Keep in mind however that like #508, we won't merge this until the next version of winit is released. We recognize that most of our users also have winit in their dependency tree and we want to avoid having multiple versions of objc2 related crates.

Thank you for your understanding.

httnn:
@DataTriny brought this PR up to speed with the current main, but seems like winit still hasn't shipped a release with the new objc2 crates?

httnn:
hmm checks are failing because the iOS adapter still uses the old objc2 crates... do we need to update the iOS adapter in this PR too?

DataTriny:
> do we need to update the iOS adapter in this PR too?

Yes, we added an iOS adapter since you opened this and it will need to be updated too. I was planning to rebase this myself once we can update winit. There are a few TODOs to address if you want to tackle it.

httnn:
updated the iOS adapter now and fixed the todos

PR #643, opened 2025-11-13
feat: Add GridCell role

lukewarlow:
Support SelectionItem Control Pattern for this new role.

A follow-up will need to implement the GridItem and TableItem Control Patterns.

This was added to Chromium in https://chromium-review.googlesource.com/c/chromium/src/+/5512762

This can be focusable, editable and selectable, but I don't think AccessKit restricts this based on role currently.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -141,6 +141,7 @@ pub enum Role {
     Footer,
     Form,
     Grid,
+    GridCell,
     Group,
     Header,
     Heading,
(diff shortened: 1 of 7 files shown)


lukewarlow:
Will rebase with main and do the cargo.toml changes once other PR is merged.

lukewarlow:
Need to remember to update this to expose this to Windows if #648 gets merged first (or vice versa)

DataTriny (approved):

PR #644, opened 2025-11-13
feat: Expose aria properties to UIA

### LINKED ISSUE #640: UIA_AriaPropertiesPropertyId property should be set for windows adaptor
Per https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-ariaspecification#w3c-aria-states-and-properties-mapped-to-microsoft-active-accessibility-and-ui-automation it seems that various states can't be neatly mapped to UIA properties and need to be handled via a fallback `UIA_AriaPropertiesPropertyId` property.

An issue with doing this is that AccessKit uses Toggled not Checked and Pressed and Selected separately, but like is done with role mapping we can probably map back to checked/pressed/selected as appropriate?

lukewarlow:
Fixes #640

### DIFF
--- a/platforms/windows/src/node.rs
+++ b/platforms/windows/src/node.rs
@@ -11,8 +11,8 @@
 #![allow(non_upper_case_globals)]
 
 use accesskit::{
-    Action, ActionData, ActionRequest, Live, NodeId, NodeIdContent, Orientation, Point, Role,
-    Toggled,
+    Action, ActionData, ActionRequest, AriaCurrent, HasPopup, Live, NodeId, NodeIdContent,
+    Orientation, Point, Role, SortDirection, Toggled,
 };
 use accesskit_consumer::{FilterResult, Node, TreeState};
 use std::sync::{atomic::Ordering, Arc, Weak};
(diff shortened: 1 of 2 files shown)


DataTriny:
I think Microsoft recommends to only include here properties or states that don't have a clear mapping to UIA. Is there added value in setting as many properties as possible?

lukewarlow:
The only one I'm slightly unsure about is readonly, because that isn't always exposed via the ValueProvider or RangeValueProvider pattern and the aria spec explicitly says to set it's value here too. https://w3c.github.io/core-aam/#ariaReadonlyTrue

That being said I'll remove it along with the others for now. We can look into whether it actually needs adding back again including afterwards.

platforms/windows/src/node.rs
```diff
@@ -312,6 +317,92 @@ impl NodeWrapper<'_> {
...
+            properties
+                .write_property("level", &level.to_string())
+                .unwrap();
+        }
+
+        if self.0.is_multiline() {
+            properties.write_bool_property("multiline", true).unwrap();
+        }
+
+        if let Some(posinset) = self.0.position_in_set() {
```
DataTriny: As said before, we already expose this property (`UIA_PositionInSetPropertyId`). Still if you think it should appear here, please use `NodeWrapper::position_in_set` because it takes care of adding one to `Node::position_in_set` which is zero-based.

DataTriny:
Unfortunately you'll have to rewrite the git history once it's ready to be merged. We want changes to the consumer crate to be in a commit with "feat:" prefix and changes to the Windows adapter in another commit with "fix:" prefix since it does not change the public API. Sorry about that, the tool we use to manage crate versions isn't always practical when multiple crates are involved.

DataTriny:
> Do I need to do the change cargo.toml files with a new line in the feat commit?

Not here. The dependent crates will just get a patch version update when their version of accesskit_consumer will be updated, which is what we want since their public API haven't changed.

PR #652, opened 2025-11-25
Replace single thiserror usage with manual trait implementations

jplatte:
I was going to send a PR updating `thiserror` to 2.0, but noticed that with the current lockfile, all dependencies are on 1.0 and it's probably not easy to update everything to 2.0 for now. I also noticed that there's only a single use of `thiserror` though which was trivial to replace with a manual implementation of `std::fmt::Display` + `std::error::Error`.

If you prefer the derive though, no hard feelings :)

### DIFF
--- a/platforms/atspi-common/src/error.rs
+++ b/platforms/atspi-common/src/error.rs
@@ -3,20 +3,31 @@
 // the LICENSE-APACHE file) or the MIT license (found in
 // the LICENSE-MIT file), at your option.
 
-#[derive(Debug, thiserror::Error)]
+use std::fmt;
+
+#[derive(Debug)]
 pub enum Error {
-    #[error("defunct")]
     Defunct,
-    #[error("unsupported interface")]
     UnsupportedInterface,
-    #[error("too many children")]
     TooManyChildren,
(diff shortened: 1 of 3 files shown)


platforms/atspi-common/src/error.rs
```diff
@@ -3,20 +3,32 @@
...
     IndexOutOfRange,
-    #[error("too many characters")]
     TooManyCharacters,
-    #[error("unsupported text granularity")]
     UnsupportedTextGranularity,
 }
 
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let msg = match self {
```
DataTriny: I'm not a big fan of unnecessarily shortening variable names, the meaning of "msg" might not be obvious to a non-English speaker. The match expression could easily be moved into the function call.

DataTriny (changes requested):
Hi @jplatte, I wanted to drop thiserror as well, so thanks for that! Just a minor comment on naming/code structure.

DataTriny (approved):
Thanks!

PR #653, opened 2025-11-25
Upgrade hashbrown, indexmap

jplatte:
The only breaking changes ([changelog](https://github.com/rust-lang/hashbrown/blob/master/CHANGELOG.md)):

> - Bump foldhash, the default hasher, to 0.2.0.
> - Replaced `DefaultHashBuilder` with a newtype wrapper around `foldhash` instead
>   of re-exporting it directly.


DataTriny:
Thanks! Would you mind moving the hashbrown dependency at the workspace level? This would make future updates much easier.

sou06912-design (commented):
Good luck

PR #656, opened 2025-12-04
Add accesskit_adapter crate. Upgrade accesskit_winit to winit 0.31.0-beta.2

### LINKED ISSUE #654: Request: Winit 0.31 support
Winit 0.31 beta has been released, and AccessKit (specifically `accesskit-winit` will need to be updated to support it).

Some notes:

- Winit 0.31 removes the generic `UserEvent` type from the `EventLoopProxy`. It's replacement is a simple `wake_up` function. The idea is that you would send your actual message through something like a channel (`std::mpsc`, or similar).
- Winit now has a type-and-trait only `winit-core` crate.

I suspect that:

- AccessKit should be able to depend on just `winit-core` (it should not need to depend on `winit`)
- Rather than trying to directly send events to the Winit `EventLoopProxy`, it should probably just accept a callback and let the user choose what they do with the events.

(I may have some time to work on this)

nicoburns:
- Fixes https://github.com/AccessKit/accesskit/issues/654

## Motivation

- Upgrade AccessKit's Winit support for Winit 0.31 (currently in beta)
- Create a cross-platform AccessKit adapter that can be used without Winit
- The non-winit cross-platform adapter would also be useful for using AccessKit with new versions of Winit before Accesskit adds support.

## Changes Made
- Create a new `accesskit_adapter` crate that abstract accross the platform adapters without depending on Winit types
- Make `accesskit_winit` a thin wrapper around `accesskit_adapter`
- Update `accesskit_winit` to Winit 0.31.0-beta.2. Notably Winit 0.31 no longer has a generic user event type. The expectation is that users of Winit will send their own events via a side channel such as a `std::mpsc` channel. I have therefore updated `accesskit_winit` to accept a generic callback.


DataTriny:
Hi @nicoburns, thank you for this. Unfortunately I don't think winit 0.31 will be released soon.

We will want to bump our MSRV once we update winit, but I'd prefer a separate commit for this. Note that I would not merge it until it is required by winit.

I am really unsure about introducing yet another crate. The subclassing adapters are a hack, so I'd prefer to keep their use limited to where they are needed. Depending on how the NSView is setup, there may even be more to do on macOS so it is dangerous to expose a generic API with this. Hopefully winit can give us the necessary hooks at some point and we can use the proper adapters for this library.

mwcampbell:
I also disagree with breaking out the winit platform backend abstraction into an `accesskit_adapter` crate. It bakes in some assumptions, like the use of the Windows and macOS subclassing adapters and the use of android-activity, that aren't universal and shouldn't be treated as such. I'd rather have those assumptions right in front of us whenever we work with winit. Hopefully we won't have to use the subclassing adapters forever. And we shouldn't automatically inherit those assumptions when we integrate AccessKit with another cross-platform windowing abstraction.

PR #668, opened 2026-01-02
feat: expose keyboard shortcuts to AT-SPI

ndarilek:


platforms/atspi-common/src/node.rs
```diff
@@ -382,6 +382,10 @@ impl NodeWrapper<'_> {
         self.0.braille_role_description()
     }
 
+    fn keyboard_shortcut(&self) -> Option<&str> {
+        self.0.data().keyboard_shortcut()
```
DataTriny: We have dedicated getters to `Node` for each property in the consumer crate, please add a new one.

DataTriny (changes requested):
Hey, thanks for that. What is the expected format though? I did not tackle this property before because I don't think we can just pass whatever string the user provided. Some platforms may require a specific format, so we have to decide which one we want from the user and apply a mapping if necessary. Implementing this property will require you to check what Chromium does. Even if you just want to implement it for the Unix adapter, you'll still have to ensure that we won't have to change the format when we come to the other platforms.

A minimal test application (pasting a modified version of the winit example in your pull request description for instance) would also be appreciated.

ndarilek:
What do you think about switching to a more structured approach for keyboard shortcuts--a struct with modifiers rather than just a plain string, similar to the recent color changes?

There do appear to be subtleties per platform. In particular Orca requires ";;<keyboard shortcut>" with a hacky fallback that presents the keyboard shortcut if it a) lacks the trailing semicolons and b) contains "ctrl".

PR #669, opened 2026-01-03
feat: Implement the `url` property

DataTriny:
I couldn't find a screen reader that reports this one. I don't know if it is useful to other ATs. Servo will likely set it for link nodes.


DataTriny:
It indeed does. Thank you @lukewarlow for teaching me how to use a screen reader! To be fair, I only use a Mac for testing purpose but couldn't find this shortcut in the documentation.

Your help is much appreciated!

mwcampbell (approved):

PR #698, opened 2026-03-16
Feat: Add UIA trees

### LINKED ISSUE #32: Add UIA support for tree views
These are like list boxes (#23), except that the items are arranged in a hierarchy, so we'll need to implement the UIA expand/collapse pattern.

amerikrainian:
As per the title, added the expanded, collapsed, and tree view item announcements for UIA. An example to test this can be found [here](https://gist.github.com/amerikrainian/632d7f85098deecbea1b2e6f84019b74)

Feedback welcome!

Closes #32

### DIFF
--- a/consumer/src/node.rs
+++ b/consumer/src/node.rs
@@ -609,8 +609,14 @@ impl<'a> Node<'a> {
         &self,
         filter: &impl Fn(&Node) -> FilterResult,
     ) -> Option<usize> {
-        self.selection_container(filter)
-            .and_then(|c| c.size_of_set())
+        let mut parent = self.filtered_parent(filter);
+        while let Some(node) = parent {
+            if let Some(size_of_set) = node.size_of_set() {
+                return Some(size_of_set);
+            }
+            parent = node.filtered_parent(filter);
+        }
+        None
     }
 
     pub fn size_of_set(&self) -> Option<usize> {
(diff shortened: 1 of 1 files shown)


platforms/windows/src/node.rs
```diff
@@ -671,10 +671,18 @@ impl NodeWrapper<'_> {
 
     fn size_of_set(&self) -> Option<i32> {
         self.0
-            .size_of_set_from_container(&filter)
+            .size_of_set()
```
DataTriny: Unlike ARIA and Chromium, the `size_of_set` AccessKit property is set on the container. As such I don't think this change is desired.

platforms/windows/src/node.rs
```diff
@@ -671,10 +671,18 @@ impl NodeWrapper<'_> {
...
+            .or_else(|| self.0.size_of_set_from_container(&filter))
             .and_then(|s| s.try_into().ok())
     }
 
+    fn level(&self) -> Option<i32> {
+        self.0
+            .level()
+            .and_then(|level| level.try_into().ok())
+            .map(|level: i32| level + 1)
```
DataTriny: Nit: by merging the map into the previous line we can guarantee that the value won't underflow.

platforms/windows/src/node.rs
```diff
@@ -687,6 +695,18 @@ impl NodeWrapper<'_> {
         self.0.supports_text_ranges()
     }
 
+    fn is_expand_collapse_pattern_supported(&self) -> bool {
+        self.0.role() == Role::TreeItem || self.0.supports_expand_collapse()
```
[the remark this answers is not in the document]
amerikrainian: This should be done now, though I must admit I am the least confident about this; I think I included everything but please let me know if that's not the case.

DataTriny:
Also please don't forget to run `cargo fmt` on your changes. Code formatting will be enforced by the CI once I trigger it for you.

DataTriny (approved):
Congratulation on your first pull request here! Thanks!

PR #707, opened 2026-04-13
feat: Allow looking up nodes by LocalNodeId and TreeId

### LINKED ISSUE #701: Consumer `NodeId` public API.
When updating AccessKit for Masonry in [xilem#1699](https://github.com/linebender/xilem/pull/1699), one of the issues we ran into is that `accesskit_consumer::NodeId` doesn't have a public API for constructing it. For the time being we can temporarily work around that with unsafe code writing directly to memory, but this isn't a suitable long term solution.

`NodeId` has three methods `new`, `with_same_tree`, and `to_components` which are currently `pub(crate)`. The easiest solution that would satisfy our needs is to change these to `pub` instead, for which I can also make the PR if that approach is acceptable.

If you have a strong reason to keep `NodeId` private to prevent construction of it, then that would also be good to know, so we could then discuss how Masonry's use of AccessKit for testing could continue in some other way.

DataTriny:
Closes #701

### DIFF
--- a/consumer/src/tree.rs
+++ b/consumer/src/tree.rs
@@ -22,13 +22,17 @@ struct TreeIndexMap {
 }
 
 impl TreeIndexMap {
-    fn get_index(&mut self, id: TreeId) -> TreeIndex {
+    fn get_or_create_index(&mut self, id: TreeId) -> TreeIndex {
         *self.id_to_index.entry(id).or_insert_with(|| {
             let tree_index = TreeIndex(self.next);
             self.next += 1;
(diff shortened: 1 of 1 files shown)


DataTriny:
cc @PoignardAzur, @XStrom

mwcampbell (approved):
Looks good to me. @DataTriny I'll let you decide whether you want to merge now or wait for feedback from the people you CC'd.

PoignardAzur:
Awesome!

Issue #565, opened 2025-05-29
iOS adapter


DataTriny:

DataTriny:
Hello @MarcelStruckWO, thank you for your interest in this.

I will be working on the iOS adapter in the coming months thanks to a grant received from the NLnet foundation, as announced yesterday.

Any help testing or reviewing code will be welcome. I will let you know once I start working on it.

iMacHumphries:
Hi @DataTriny, really glad to see this feature is on the way! I wanted to introduce myself as a downstream user in case it's useful for prioritization or testing.

I'm developing a 2D mobile game in the Godot Engine, which already integrates AccessKit on macOS, Windows, and Linux.  Once accesskit_ios is available, I'd like to wire it into Godot's iOS DisplayServer and try it out in my game. My game is essentially all 2D menus, so it would be a near-ideal validation target for the iOS adapter once it lands.   You can check it out here: https://heroiclegends.net

A couple questions to help me plan, and see if I can help out.
1. Timeline.  I saw the commit in January and the 4 phase plan in this issue.  Do you have a rough sense of when Phase 1 might be testable / when phase 4 would follow?   I'm trying to decide if I should wait for upstream or build a temp UIKit accessibility backend for my Godot fork.
2. Where can I help? You mentioned testing and code review would be appreciated. I'd be glad to test pre-release builds in Godot.   I'm also interested in contributing if that can help speed this adaptor along.

No rush on a reply, but I just wanted to put a real downstream product on your radar to see if that helps with prioritization, and so I can plan our roadmap.  Thanks for taking this on!

DataTriny:
Hi @iMacHumphries, thanks for your message.

Phase 1 is ~80% complete and I plan to open the pull request in the coming days. Integration in our C bindings was planned in Phase 4 as I didn't expect non-Rust users to take advantage of it so soon. Your game could be a very nice showcase though, I can easily add the 0.1 version of the iOS adapter there. The path for Phase 3 is a bit uncertain at the moment and might require additions to the C API to support editable text. Is this something you would need right away?

iMacHumphries:
@DataTriny This is so awesome, thank you! A 0.1 version in the C bindings would be exactly what I need to start adding this to Godot / Heroic Legends.

[...]

PR #710, opened 2026-04-19
feat: Basic iOS adapter

DataTriny:
Phase 1 of #565 

Implements the adapter for iOS and other Apple systems that use the UIKit runtime.

The adapter comes in two flavours: a low-level adapter suitable when the user owns the UIView and a subclassing adapter useful when integrating with third-party libraries that don't let the user manipulate the UIView, such as winit.

This first version is able to expose simple widgets like labels, buttons, checkboxes, switches, sliders... Live regions are also supported.

Necessary infrastructure is in place to run tests on a simulator, the CI is modified to take advantage of this and run the unit tests.

accesskit_winit now use the iOS subclassing adapter on appropriate targets. The examples have been modified and can be used to manually test the adapter.

The Slint toolkit is already compatible and its Gallery example can be used for further manual testing, although issues in its layout make VoiceOver refuse to focus certain widgets. Please use [this branch](https://github.com/DataTriny/slint/tree/ios) where the AccessKit crates are patched.

Editable text is not yet implemented and focusing such a widget can confuse VoiceOver, depending on how the view is set up.


mwcampbell:
Do we want to clear `Context::platform_focus` if the node that that field refers to is destroyed?

mwcampbell:
@DataTriny Can you please add a comment in the winit examples explaining the reason for the announcement delay?

mwcampbell (approved):
This is a good, thorough implementation of phase 1. The delay isn't ideal, but the better solution, if there is one, isn't obvious, so I won't block on that.

platforms/ios/src/util.rs
```diff
@@ -0,0 +1,57 @@
...
+unsafe impl Encode for UIAccessibilityExpandedStatus {
+    const ENCODING: Encoding = NSInteger::ENCODING;
+}
+
+unsafe impl RefEncode for UIAccessibilityExpandedStatus {
+    const ENCODING_REF: Encoding = Encoding::Pointer(&Self::ENCODING);
+}
+
+pub(crate) fn from_cg_point(view: &UIView, node: &Node, point: CGPoint) -> Point {
+    let local_point = unsafe { view.convertPoint_fromView(point, None) };
```
iMacHumphries: The coordinate conversion uses `convertPoint_fromView(point, None)` / `convertRect_toView(rect, None)`, which converts to/from window coordinates. [accessibilityFrame](https://developer.apple.com/documentation/uikit/uiaccessibilityelement/accessibilityframe) is expected to return screen coordinates. On fullscreen iPhone apps these are the same, but on iPad with Split View the window doesn't fill the screen, so VoiceOver could get incorrect element positions. 

You should be able to use this to convert to screen space:
https://developer.apple.com/documentation/uikit/uiaccessibility/converttoscreencoordinates(_:in:)-9ziiu

iMacHumphries: Also, we might not need to handle safe area here.  If a view's frame is within the safe area, then we'd want the accessibilityFrame to match the view's frame exactly.

DataTriny: I went back and forth on how to handle the safe area. Ultimately I think you're right, the adapter should not guess it. This adds a bit more complexity to our examples but that's really something every iOS app would have to  do anyway.

DataTriny:
I am happy with the current state of this pull request. I am going to merge this tomorrow if there are no objections.

DataTriny:
Actually I noticed that VoiceOver takes way more time to focus the first node of an AccessKit app than it does for a stock app. I want to see if this is related to the adapter initialization being fully dependent on VoiceOver making the first query.

DataTriny:
It is indeed very slow (~1.5-2 seconds) to wait for VoiceOver to perform its first query on a view. I have implemented essentially what Flutter does: observe the ATs and check their current state as soon as possible to decide whether to activate the adapter. With that infrastructure in place we can support deactivation. The old path is kept for third-party ATs.

The examples had to be modified again because we weren't applying the scale factor. This caused obvious positionning issues on landscape orientation. As a bonus they now demonstrate the good pattern of applying a transform to the root.

mwcampbell (approved):
The latest changes look good, particularly the new approach to activation and deactivation.

PR #715, opened 2026-05-22
fix: Watch the IsEnabled property to lazily activate on Unix

### LINKED ISSUE #713: AccessKit no longer detects when Orca is running
I'm not sure how to properly test AccessKit itself on this, but since a certain point, frameworks using accessKit seem no longer able to detect Orca, and in turn, the a11y support is not turned on. eframe 0.34.1, Slint 1.16 and Vizia 0.4 from what I've tested are all not accessible on Linux currently. The older version of these frameworks, e.g. Slint 1.13 seem to work correctly.

## Environment

openSUSE Tumbleweed from start of this month, Gnome 49 Wayland, Orca 51.alpha (tested with 49.5 as well, same result).

DataTriny:
Closes #713 

Following [this commit in at-spi2-core](https://gitlab.gnome.org/GNOME/at-spi2-core/-/commit/c674dc9740ca3858cd90f4206d2ee4309fba7388) we need to stop relying on the now removed `ScreenReaderEnabled` bus property and switched to monitor `IsEnabled` instead, which should be more reliable.

### DIFF
--- a/platforms/unix/src/context.rs
+++ b/platforms/unix/src/context.rs
@@ -21,7 +21,7 @@ use tokio::{
 };
 #[cfg(feature = "tokio")]
 use tokio_stream::{StreamExt, wrappers::UnboundedReceiverStream};
-use zbus::{Connection, connection::Builder};
+use zbus::{Connection, connection::Builder, proxy::PropertyChanged};
 
 use crate::{
     adapter::{AdapterState, Callback, Message},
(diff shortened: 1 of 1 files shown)


DataTriny:
I'm not currently able to validate this on a Freedesktop environment.

@RastislavKish it would be great if you could apply [lazy-activation-debug.patch](https://github.com/user-attachments/files/28164065/lazy-activation-debug.patch) on top of this branch then try the simple winit example. Let me know if the tree gets initialized immediately when Orca was previously running, if it can be first initialized later if Orca is not initially running and if it can be deactivated and reactivated during the app lifecycle.

Thank you for your help.

RastislavKish:
@DataTriny looks good! I have installed at-spi-2-core from source and things seem to be working well, I can see both activation and deactivation reports as I turn on / off the screenreader. Everything also works as expected regardless when the screenreader is started i.e. before or after launching the app. Great work on both your and Mike's side!

mwcampbell (approved):
@DataTriny Looks good to me, and I see we got a successful test result. Please merge and release when ready.

PR #717, opened 2026-05-25
refactor: Make consumer tree traversals iterative

DataTriny:
Using recursive functions on trees whose depth is controlled by the user is dangerous as it can lead to stack overflows.

Increases test coverage of affected functions before making them iterative. We might want to bring smallvec in the future to avoid some memory allocations.


mwcampbell (approved):
Looks good to me, especially with the extra test coverage. Feel free to merge when ready.

PR #718, opened 2026-06-02
fix: Compute index in parent for tree roots on Unix

DataTriny:
When computing the index in parent for the root node of an adapter, return the index of the adapter in the list of adapters. Otherwise all such nodes would report 0 as their index.


mwcampbell (approved):
LGTM. Please merge when ready. I think I want to switch to having the author merge after an approving review, like Linebender does.

PR #719, opened 2026-06-05
feat: Implement the cache object on Unix

DataTriny:
AT-SPI expects all applications to expose a cache object which libatspi uses to populate its tree. libatspi outputs a warning if it can't find it, which make AccessKit looks bad.


mwcampbell (approved):
Looks good to me. I especially appreciate the continued focus on adding more tests.

PR #721, opened 2026-06-09
Shorten some type signatures returning `impl DoubleEndedIterator`

PoignardAzur:
- Remove `FusedIterator` bound, which was redundant. (User code never needs to bind on `FusedIterator`, that trait is only used by the standard library for specialization.)
- Remove redundant `+ 'a` bounds.
- Remove redundant `<Item = ...>` parameter from `ExactSizeIterator` bounds (there's no need to specify parameters twice).

### DIFF
--- a/consumer/src/node.rs
+++ b/consumer/src/node.rs
@@ -17,7 +17,7 @@ use alloc::{
     vec,
     vec::Vec,
 };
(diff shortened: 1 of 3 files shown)


DataTriny (approved):
Thanks!

PR #729, opened 2026-06-22
Allow graceful symbol degration to support iOS 15+ at runtime

molenick:
First off - congrats on the release of accesskit_ios! 🎉

I started the process of using accesskit_ios to expose accessibility metadata in my egui app and was getting an immediate runtime crash on iOS 15.8.8. I looked into it and found that a few iOS 17+ symbols were being unconditionally loaded. This work gates the loading of those symbols on iOS 17+ so that iOS 15+ apps can make use of most of accesskit_ios's features.

Testing/Verification
In my private repo I was able to pin accesskit_ios to a local copy of this PR and verify that app was able to build, run and receive accesskit metadata on a real iPhone SE 1st gen running iOS 15.8.8

### DIFF
--- a/platforms/ios/src/util.rs
+++ b/platforms/ios/src/util.rs
@@ -6,10 +6,10 @@
 use accesskit::Point;
 use accesskit_consumer::Node;
 use objc2::encode::{Encode, Encoding, RefEncode};
-use objc2_foundation::{CGPoint, CGRect, CGSize, NSInteger};
+use objc2_foundation::{CGPoint, CGRect, CGSize, NSAttributedStringKey, NSInteger, NSString};
 use objc2_ui_kit::{
-    UIAccessibilityConvertFrameToScreenCoordinates, UIAccessibilityTraits, UICoordinateSpace,
-    UIView,
+    UIAccessibilityConvertFrameToScreenCoordinates, UIAccessibilityPriority, UIAccessibilityTraits,
+    UICoordinateSpace, UIView,
 };
 use std::ffi::{c_char, c_void};
 use std::sync::OnceLock;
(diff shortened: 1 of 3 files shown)


platforms/ios/src/util.rs
```diff
@@ -56,3 +61,21 @@ pub(crate) fn to_cg_rect(view: &UIView, rect: accesskit::Rect) -> CGRect {
     };
     to_screen_rect(view, local_rect)
 }
+
+const RTLD_DEFAULT: *mut c_void = -2isize as *mut c_void;
```
DataTriny, line 71: What's this constant? It might be worth putting a comment saying where it come from.

DataTriny (changes requested):
The market share of iOS <17 seem very small so I don't think I want to put much effort into that, but this is reasonable.

Looks good, just one comment.

platforms/ios/src/util.rs
molenick: Good catch! I added a comment to explain its usage with dlsym (https://github.com/AccessKit/accesskit/pull/729/commits/f1dedfb95e85dffe5933b7345d92e03407379930) and added links to documentation and header sources. Thanks for consideration, let me know if anything else needs attention.

DataTriny (approved):
Thank you! Supporting iOS 15+ seem reasonable for now and we will try to stick to it for a while. This is useful information that should be communicated somewhere. I would welcome a pull request adding this where you would have expected to find this information instead of discovering it at runtime. I will do it myself later otherwise.

DataTriny:
Forgot to edit the commit message before merging so the crate version and changelog gets updated. Rewriting the commit on the main branch was unfortunately the better option. @molenick This is described in our [contributing guide](https://github.com/AccessKit/accesskit/blob/main/CONTRIBUTING.md), please refer to it for your future contributions.

PR #731, opened 2026-06-24
Consumer Coverage and a11y Flag Wiring

William-Selna:
## In this PR
7 commits - I originally planned on purely bumping function test coverage, but I found an odd twist in the wiring and a few stubs from 2022. At first, I intended these to be largely conformance tests (as in look at the asserted behavior), but I was skeptical of the insertion and decided to investigate.

The test I'm referring to is page_navigation_resolves_to_document_bounds(), a test to assert the following 3 position methods:
- forward_to_page_start
- forward_to_page_end
- backward_to_page_start

As was originally written, the test I wrote asserted page-start == page-end == document_end, which appears to me as counterintuitive. So I decided to look further. Pulling up the blame, it appeared as an unwired artifact from 2022. After noticing these stubs, I ran a check I'd written for some of my other projects that flags write-only state, ones settable by a provider, but not read by anything.

The test/grep flagged `is_page_breaking_object` as write-only, which is the exact flag these methods would need to consult. That explains the four stubs below. With nothing reading the flag, they could only ever resolve to document bounds.

```Rust
pub fn forward_to_page_start(&self) -> Self { self.document_end() }
pub fn forward_to_page_end(&self)   -> Self { self.document_end() }
pub fn backward_to_page_start(&self) -> Self { self.document_start() }
pub fn is_page_start(&self) -> bool  { self.is_document_start() }
// no is_page_end, no backward_to_page_end
```
## Open Decision Pending: `is_page_breaking_object`

The schema documents `is_page_breaking_object` only as "Indicates whether this node causes a page break," and no provider sets it today, so nothing pins down the tree shape a provider is expected to emit. Implementing detection requires
committing to a contract, and I assumed: **A node carrying `is_page_breaking_object` is a container wrapping a page's
content, and the break occurs *before* it** — so the node's first text-run descendant begins a new page (`starts_new_page` walks up from a run to find such an ancestor.)

Alternatives I did *not* assume: break-*after* (the flagged node ends a page), or the flag on a marker/separator node between pages — cf. `is_line_breaking_object`, documented as "block level elements, or `<br>`," which spans both readings.

If you intended a different contract, only the detection predicate changes, and the navigation loops and tests are unaffected. The implementation is on the branch if you want it; the contract choice (break-before / break-after / marker-node) is yours, and if you'd rather not commit to page semantics this cycle, the cleanest answer is to drop these commits and keep the rest. I'm also happy to split the PR up, I know the surface is on the larger side. 

## The Fixes for the Stubs:
- is_page_start → now true at the document start or at a line-start that begins a new page (detected by walking up to an ancestor flagged is_page_breaking_object).
- is_page_end → new method, mirroring is_paragraph_end.
forward_to_page_start / forward_to_page_end / backward_to_page_start → the one-line stubs became real loops that walk line-by-line to the next page boundary, structurally identical to the existing paragraph-nav methods.
- Added the detection helpers (starts_new_page, next_run_starts_new_page, node_contains).
- Deliberately did not add backward_to_page_end — no text unit in the crate has a backward-to-end method, so adding one would make page the inconsistent unit.
- Backward-compatible: with no is_page_breaking_object flags anywhere (every document today), all of these collapse back to exactly the old document-bounds behavior — so nothing regresses; only trees that set the flag get the new behavior.
- The methods now consume the is_page_breaking_object schema flag, which previously existed but was read by nothing.

## Wiring other write-only flags. 

The same unwired-detector run surfaced more schema flags that were set-able but read by nothing. Three had a real platform home, so this PR wires them:

- is_visited → AT-SPI State::Visited (already-followed links now announce as visited).
- is_spelling_error / is_grammar_error → text attributes: AT-SPI invalid = "spelling"/"grammar", and the Windows UIA AnnotationTypes attribute.
macOS arms are deferred — its text-attribute path is about to be reworked by the pending objc2 upgrade ([#616](https://github.com/AccessKit/accesskit/pull/616)).

## Tests Written to Assert new Behavior
**`consumer/src/text.rs` (13)**

Page navigation:
- `is_page_start_tracks_breaks_not_paragraphs`
- `forward_to_page_start_skips_intra_page_paragraphs`
- `forward_to_page_end_lands_on_page_boundaries`
- `backward_to_page_start_walks_pages_in_reverse`
- `without_page_breaks_navigation_resolves_to_document_bounds`

Spelling / grammar (auto-generated by `inherited_flags!`):
- `is_spelling_error::directly_set`
- `is_spelling_error::mixed`
- `is_spelling_error::set_on_parent`
- `is_spelling_error::unset`
- `is_grammar_error::directly_set`
- `is_grammar_error::mixed`
- `is_grammar_error::set_on_parent`
- `is_grammar_error::unset`

**`platforms/atspi-common/src/text_attributes.rs` (1)**

- `invalid_attribute_reflects_error_flags`

**`platforms/atspi-common/src/node.rs` (1)**

- `visited_flag_maps_to_atspi_visited_state`

## Overall **accesskit_consumer** Coverage improvement
File | Line	 | Function
-- | -- | --
node.rs  |	79.2% → 96.7%  |	57.4% → 96.0%
text.rs   |	90.6% → 96.4%  |	82.1% → 97.7%
tree.rs   |	91.2% → 93.0%  |	70.2% → 74.5%
Total | 89.6% → 95.6% | 72.9% → 90.1%

## Notes

The Windows AnnotationTypes arm is type-checked against x86_64-pc-windows-msvc but I have no Windows host, so its runtime behavior relies on CI. 

I can also split this PR up into just tests or however you like, please don't hesitate to let me know.

### DIFF
--- a/consumer/src/node.rs
+++ b/consumer/src/node.rs
@@ -2089,4 +2089,335 @@ mod tests {
                 .is_focused()
         );
     }
+
+    // The numeric, scroll, and semantic-state accessors are thin reads of the
+    // underlying node data, but none of them were exercised by the existing
+    // suite; these pin the setter-to-getter round trip the platform adapters
+    // depend on.
+    #[test]
+    fn numeric_and_scroll_accessors() {
+        let update = TreeUpdate {
+            nodes: vec![
+                (NodeId(0), {
+                    let mut node = Node::new(Role::RootWebArea);
+                    node.set_children(vec![NodeId(1)]);
+                    node
+                }),
+                (NodeId(1), {
+                    let mut node = Node::new(Role::Slider);
+                    node.set_numeric_value(3.0);
+                    node.set_min_numeric_value(0.0);
+                    node.set_max_numeric_value(10.0);
+                    node.set_numeric_value_step(1.0);
+                    node.set_numeric_value_jump(2.0);
+                    node.set_scroll_x(5.0);
+                    node.set_scroll_x_min(0.0);
+                    node.set_scroll_x_max(50.0);
+                    node.set_scroll_y(6.0);
+                    node.set_scroll_y_min(0.0);
+                    node.set_scroll_y_max(60.0);
+                    node
+                }),
+            ],
+            tree: Some(Tree::new(NodeId(0))),
+            tree_id: TreeId::ROOT,
+            focus: NodeId(0),
+        };
+        let tree = crate::Tree::new(update, false);
+        let state = tree.state();
+        let node = state.node_by_id(nid(NodeId(1))).unwrap();
+        assert_eq!(Some(3.0), node.numeric_value());
+        assert_eq!(Some(0.0), node.min_numeric_value());
+        assert_eq!(Some(10.0), node.max_numeric_value());
+        assert_eq!(Some(1.0), node.numeric_value_step());
+        assert_eq!(Some(2.0), node.numeric_value_jump());
+        assert_eq!(Some(5.0), node.scroll_x());
+        assert_eq!(Some(0.0), node.scroll_x_min());
+        assert_eq!(Some(50.0), node.scroll_x_max());
+        assert_eq!(Some(6.0), node.scroll_y());
+        assert_eq!(Some(0.0), node.scroll_y_min());
+        assert_eq!(Some(60.0), node.scroll_y_max());
+    }
+
+    #[test]
+    fn state_and_semantic_accessors() {
+        let update = TreeUpdate {
+            nodes: vec![
+                (NodeId(0), {
+                    let mut node = Node::new(Role::RootWebArea);
+                    node.set_children(vec![NodeId(1)]);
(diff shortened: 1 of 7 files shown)


DataTriny:
Hello @William-Selna, thank you for your interest in helping here.

Unfortunately this pull request contain too much and barely related changes. Especially important for a first time contributor, please focus on contained changes that will be easier to review.

Most of your tests also do too much, prefer smaller unit tests that would be easier to understand if they break. Testing very simple methods for the sake of increasing coverage is not a priority either.

When exposing node properties through the platform adapters, remember that Chromium is our reference implementation. Try to handle as much platforms as possible, and link a test program using the accesskit_winit crate to your pull request, so we can at least perform manual testing.

I am going to keep this pull request open for now, letting you decide how you want to re-scope it. But I won't review it as is. Thank you for your understanding.

PR #738, opened 2026-06-28
refactor: Reuse buffers for wide string conversion on Windows

DataTriny:
Initially extracted from [stacked-alloc-refactors](https://github.com/accesskit/accesskit/tree/stacked-alloc-refactors) but significantly reworked.

Introduce reusable buffers for wide string conversion in the Windows adapter to reduce memory allocations. The initial work used two separate buffers tied to the adapter instance, I chose to go with a thread-local pool of buffers, mostly to avoid acquiring a lock for each node property getter, even if they don't convert strings. Multiple adapters can also share the same buffers. I capped the buffer size to avoid keeping large memory chunks around.


mwcampbell (approved):
Thanks for taking the time to improve on my version and add test coverage.

PR #739, opened 2026-07-03
Add `cargo doc` pass to Github Actions

PoignardAzur:
Copy-pasted from Vello's CI script.


DataTriny:
Have you missed #736? Most notable difference seem to be the toolchain, but we usually don't care about nightly that much here.

PoignardAzur:
Oh, I forgot to update my main branch. My bad.

PR #740, opened 2026-07-05
fix: Set the desktop as the app's parent on Unix

DataTriny:
Stores the AT-SPI desktop plug returned by `Socket.Embed` somewhere shared by the root accessible node and the cache, so that both can return it when asked.

This was dropped as part of a refactoring a while back because GTK4 didn't bother and it didn't seem to matter. A recent change in the Orca screen reader make this mandatory now.


DataTriny:
@ndarilek I'd appreciate if you could try one of our winit examples with and without this patch, with the newest Orca version. It should fix the window focusing issue.

Thanks.

ndarilek:
Yes, seems to work. Thanks for the fix!

PR #741, opened 2026-07-07
fix: Don't panic on recoverable D-Bus errors on Unix

### LINKED ISSUE #727: A failed AT-SPI connection panics (unwrap in run_event_loop)
On Linux, `accesskit_unix` connects to the AT-SPI bus on a background thread. If that connection fails for any reason, the error is `unwrap`'d:

```rust
// src/context.rs, get_or_init_messages()
run_event_loop(&executor, session_bus, rx).await.unwrap();
```

With `panic = "abort"`, that aborts the whole process, so the app crashes on startup just because an optional accessibility connection failed. The inner error varies by environment (`Handshake("Server GUID mismatch …")`, `NotFound`, or `InterfaceNotFound`.

This became easy to hit in accesskit_unix 0.22.0, which changed the gate from `ScreenReaderEnabled` to `IsEnabled`:

```rust
// 0.21.1
let changes = status.receive_screen_reader_enabled_changed().await.fuse();
// 0.22.0
let changes = status.receive_is_enabled_changed().await.fuse();
```

`IsEnabled` is true on a normal desktop even with no screen reader, so AccessKit now connects for nearly every app, and any unreachable AT-SPI bus crashes it.

In my case, I observed this when cargo update upgraded to accesskit 0.33.1 and therefore accesskit_unix 0.22.
As a developer i might have an unconventional setup and i do run several sessions for tests and things and got panic on all my slint apps ``panicked at accesskit_unix-0.22.0/src/context.rs:61:78:
called `Result::unwrap()` on an `Err` value: Handshake("Server GUID mismatch: expected 7ed5678e88c8df9526ef139d6a1b354e, got 0c1e09f80bf08973a2020b806a366325")``

### Reproduce

```sh
# a bus that doesn't exist (misconfiguration)
export AT_SPI_BUS_ADDRESS="unix:path=/tmp/no-such-a11y-bus,guid=00000000000000000000000000000000"

cargo run -p accesskit_winit --example simple
```

```
thread '<unnamed>' panicked at accesskit_unix-0.22.0/src/context.rs:61:
called `Result::unwrap()` on an `Err` value:
InputOutput(Os { code: 2, kind: NotFound, message: "No such file or directory" })
```

### Expected

A failed AT-SPI connection should be logged and accessibility disabled for the session; the app keeps running.

DataTriny:
Closes #727 

A system misconfiguration, a broken AT-SPI registry implementation or an adversarial assistive technology should never make the host application crash. This pull request ensure only a bug on our side can lead to a panic.

### DIFF
--- a/platforms/unix/src/atspi/bus.rs
+++ b/platforms/unix/src/atspi/bus.rs
@@ -18,7 +18,6 @@ use atspi::{
 };
 use std::{
     env::var,
-    io,
     sync::{Arc, OnceLock},
 };
 use zbus::{
@@ -172,7 +171,7 @@ impl Bus {
     where
         T: zbus::object_server::Interface,
     {
-        map_or_ignoring_broken_pipe(
+        map_or_ignoring_recoverable_error(
             self.conn.object_server().at(path, interface).await,
             false,
             |result| result,
@@ -223,7 +222,7 @@ impl Bus {
     where
         T: zbus::object_server::Interface,
     {
-        map_or_ignoring_broken_pipe(
+        map_or_ignoring_recoverable_error(
             self.conn.object_server().remove::<T, _>(path).await,
             false,
             |result| result,
@@ -415,7 +414,7 @@ impl Bus {
     where
         B: serde::Serialize + zbus::zvariant::DynamicType,
     {
-        map_or_ignoring_broken_pipe(
+        map_or_ignoring_recoverable_error(
             self.conn
                 .emit_signal(
                     Option::<BusName>::None,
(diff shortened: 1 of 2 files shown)


DataTriny:
@ogoffart Could you please confirm that this doesn't crash a Slint application anymore on your broken AT-SPI setup?

I haven't thought too much about the tracing/logging story for AccessKit yet and this needs to be the same for all adapters, so that will come later.

Thanks.

ogoffart:
Thanks!

Yes, this solves the panic i had.

mwcampbell (approved):
LGTM. @DataTriny Please merge when ready.

PR #744, opened 2026-07-13
fix: Address Rust 1.97 clippy warnings

DataTriny:


PoignardAzur (approved):

PoignardAzur:
CI is failing because of the `quick-xml` advisory. Pretty sure it doesn't concern us.

Also, I don't know how configurable `cargo-deny-action` is, but I don't think we should block CI over denial-of-service vulnerabilities.

PR #746, opened 2026-07-13
fix: Update zbus_xml to 5.2

DataTriny:
This removes one of two paths through which we depend on quick-xml: `cargo update -p zbus_xml`

- [RUSTSEC-2026-0194](https://rustsec.org/advisories/RUSTSEC-2026-0194)
- [RUSTSEC-2026-0195](https://rustsec.org/advisories/RUSTSEC-2026-0195)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -995,9 +995,9 @@ dependencies = [
 
 [[package]]
 name = "memchr"
-version = "2.5.0"
+version = "2.8.3"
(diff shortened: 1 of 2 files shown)


PoignardAzur (approved):

DataTriny:
Unsure if we actually want to go with it, we'd have to bump our MSRV first. We might ignore this vulnerability path at least for one release. XML files are only read at build time by zbus-lockstep.

Issue #749, opened 2026-07-14
Document accesskit properties and roles that don't match ARIA


PoignardAzur:
Accesskit types are ported for chromium's API, which don't always match ARIA properties.

I've gone through the entire property list with @DataTriny and we wrote the mismatched properties, though my notes are sparse and I usually didn't note what the exact mismatch was. Still, if we want to document our relationship with ARIA, these properties are a good starting point:

- IsSpellingError
- IsGrammarError
- IsSearchMatch
- IsSuggestion
- Children
- InPageLinkTarget
- MemberOf
- NextOnLine
- PreviousOnLine
- PopupFor
- Label
- Value
- ClassName
- HtmlTag
- Url
- RowIndexText
- ColumnIndexText
- ScrollX
- ScrollXMin
- ScrollXMax
- ScrollY
- ScrollYMin
- ScrollYMax
- NumericValue
- NumericValueStep
- NumericValueJump
- FontSize
- FontWeight
- BackgroundColor
- ForegroundColor
- Overline
- Strikethrough
- Underline
- Invalid
 - Toggled
- TextDirection
- ListStyle
- TextAlign
- VerticalOffset
- TextSelection

Ideally we should also document differences in roles and actions, but I haven't started to list those.

PoignardAzur:
Some additional notes:

- Some properties are one-indexed in ARIA and zero-indexed in AccessKit, such as RowIndex, ColumnIndex, Level, and PositionInSet.
- Label is equivalent to `name` in ARIA.
- Value is somewhat equivalent to `valuetext` in ARIA.
- NumericValue is equivalent to `valuenow` in ARIA.
- MinNumericValue is equivalent to `valuemin` in ARIA.
- MaxNumericValue is equivalent to `valuemax` in ARIA.
- Toggled is somewhat equivalent to either `checked` or `pressed` in ARIA.

blinding-pixels:
I would be happy to start on this. I think maybe starting with a bounded scope on a smaller subsection could be good. That way we can nail down the format, language and tone?

I want to make sure I understand correctly: for each AccessKit role, property, etc., we want to explain what it means, link to its equivalent ARIA item if one exists, and note any differences from the ARIA version?

Putting down an example below as per my current understanding.

## `Role::Button`

### AccessKit definition
`Role::Button` represents a control that performs an action when activated. The node should normally have an accessible label and support `Action::Click`.

### ARIA equivalent
[...]

PR #753, opened 2026-07-20
Document ARIA mappings for string properties

blinding-pixels:
This is a first pass at #749, starting with the string properties as discussed.

The inline documentation covers the following -:

- explains what each string property represents;
- links to the corresponding ARIA concept or property where one exists;
- notes when there is no direct ARIA equivalent; and
- calls out specific differences for `Value` and `KeyboardShortcut`.

The newer row and column index text and Braille properties link to the WAI-ARIA 1.3 draft and are identified as draft equivalents.

This is intentionally limited to one property type so we can agree on the format and wording before expanding the documentation further.

I ran the following checks,-

- `cargo test -p accesskit`
Happy to revise anything if required.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -1909,61 +1909,114 @@ node_id_property_methods! {
 }
 
 string_property_methods! {
-    /// The label of a control that can have a label. If the label is specified
-    /// via the [`Node::labelled_by`] relation, this doesn't need to be set.
-    /// Note that the text content of a node with the [`Role::Label`] role
-    /// should be provided via [`Node::value`], not this property.
+    /// The accessible name of this node. If the name is specified via the
+    /// [`Node::labelled_by`] relation, this doesn't need to be set. The text
+    /// content of a node with the [`Role::Label`] role should be provided via
+    /// [`Node::value`], not this property. ARIA equivalent: [accessible name].
+    ///
+    /// [accessible name]: https://www.w3.org/TR/wai-aria-1.2/#dfn-accessible-name
     (Label, label, set_label, clear_label),
+    /// A short explanation that supplements this node's accessible name.
+    /// ARIA equivalent: [accessible description].
+    ///
+    /// [accessible description]: https://www.w3.org/TR/wai-aria-1.2/#dfn-accessible-description
     (Description, description, set_description, clear_description),
+    /// The text content or string value of this node. ARIA equivalent:
+    /// [`aria-valuetext`] for range widgets.
+    ///
+    /// **Difference with ARIA:** Unlike `aria-valuetext`, this property is also
+    /// used for text controls and the text content of nodes such as [`Role::Label`].
+    ///
+    /// [`aria-valuetext`]: https://www.w3.org/TR/wai-aria-1.2/#aria-valuetext
     (Value, value, set_value, clear_value),
     /// A single character, usually part of this node's name, that can be pressed,
     /// possibly along with a platform-specific modifier, to perform
     /// this node's default action. For menu items, the access key is only active
     /// while the menu is active, in contrast with [`keyboard_shortcut`];
-    /// a single menu item may in fact have both properties.
+    /// a single menu item may in fact have both properties. ARIA has no distinct
+    /// equivalent; [`aria-keyshortcuts`] corresponds to [`keyboard_shortcut`].
     ///
     /// [`keyboard_shortcut`]: Node::keyboard_shortcut
+    /// [`aria-keyshortcuts`]: https://www.w3.org/TR/wai-aria-1.2/#aria-keyshortcuts
     (AccessKey, access_key, set_access_key, clear_access_key),
     /// A way for application authors to identify this node for automated
-    /// testing purpose. The value must be unique among this node's siblings.
+    /// testing purposes. The value must be unique among this node's siblings.
+    /// There is no ARIA equivalent.
     (AuthorId, author_id, set_author_id, clear_author_id),
+    /// The platform control class name for this node. There is no ARIA equivalent.
     (ClassName, class_name, set_class_name, clear_class_name),
-    /// Only present when different from parent.
+    /// The font family used for this node's text. Only set this when it differs
+    /// from the parent. There is no ARIA equivalent.
     (FontFamily, font_family, set_font_family, clear_font_family),
+    /// The name of the HTML element represented by this node. There is no ARIA
+    /// equivalent.
     (HtmlTag, html_tag, set_html_tag, clear_html_tag),
     /// Inner HTML of an element. Only used for a top-level math element,
     /// to support third-party math accessibility products that parse MathML.
+    /// There is no ARIA equivalent.
     (InnerHtml, inner_html, set_inner_html, clear_inner_html),
     /// A keystroke or sequence of keystrokes, complete with any required
     /// modifiers(s), that will perform this node's default action.
-    /// The value of this property should be in a human-friendly format.
+    /// The value of this property should be in a human-friendly format. ARIA
+    /// equivalent: [`aria-keyshortcuts`].
+    ///
+    /// **Difference with ARIA:** `aria-keyshortcuts` uses a standardized syntax,
+    /// while AccessKit expects a human-friendly string.
+    ///
+    /// [`aria-keyshortcuts`]: https://www.w3.org/TR/wai-aria-1.2/#aria-keyshortcuts
     (KeyboardShortcut, keyboard_shortcut, set_keyboard_shortcut, clear_keyboard_shortcut),
     /// An [IETF language tag](https://www.rfc-editor.org/info/bcp47).
-    /// Only present when different from parent.
+    /// Only set this when it differs from the parent. There is no ARIA equivalent;
+    /// web content supplies this information through the host language.
     (Language, language, set_language, clear_language),
     /// If a text input has placeholder text, it should be exposed
-    /// through this property rather than [`label`].
+    /// through this property rather than [`label`]. ARIA equivalent:
+    /// [`aria-placeholder`].
     ///
     /// [`label`]: Node::label
+    /// [`aria-placeholder`]: https://www.w3.org/TR/wai-aria-1.2/#aria-placeholder
     (Placeholder, placeholder, set_placeholder, clear_placeholder),
     /// An optional string that may override an assistive technology's
     /// description of the node's role. Only provide this for custom control types.
-    /// The value of this property should be in a human-friendly, localized format.
+    /// The value should be human-friendly and localized. ARIA equivalent:
+    /// [`aria-roledescription`].
+    ///
+    /// [`aria-roledescription`]: https://www.w3.org/TR/wai-aria-1.2/#aria-roledescription
     (RoleDescription, role_description, set_role_description, clear_role_description),
     /// An optional string that may override an assistive technology's
     /// description of the node's state, replacing default strings such as
     /// "checked" or "selected". Note that most platform accessibility APIs
-    /// and assistive technologies do not support this feature.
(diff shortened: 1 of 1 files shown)


common/src/lib.rs
```diff
@@ -1909,61 +1909,114 @@ node_id_property_methods! {
 }
 
 string_property_methods! {
-    /// The label of a control that can have a label. If the label is specified
-    /// via the [`Node::labelled_by`] relation, this doesn't need to be set.
-    /// Note that the text content of a node with the [`Role::Label`] role
-    /// should be provided via [`Node::value`], not this property.
+    /// The accessible name of this node. If the name is specified via the
```
DataTriny: We intentionally renamed this property from "name" to "label" because we thing it is more understandable for UI toolkit authors, please keep it this way. The ARIA equivalent here should be `aria-label` anyway. Also, I think the word "accessible" anywhere in the documentation is just noise, as it's AccessKit's mission.

common/src/lib.rs
```diff
@@ -1909,61 +1909,114 @@ node_id_property_methods! {
...
+    /// [`Node::labelled_by`] relation, this doesn't need to be set. The text
+    /// content of a node with the [`Role::Label`] role should be provided via
+    /// [`Node::value`], not this property. ARIA equivalent: [accessible name].
+    ///
+    /// [accessible name]: https://www.w3.org/TR/wai-aria-1.2/#dfn-accessible-name
     (Label, label, set_label, clear_label),
+    /// A short explanation that supplements this node's accessible name.
```
DataTriny: By definition a description is not "short". The ARIA equivalent would be `aria-description`. You should reference "label" instead of "accessible name".

common/src/lib.rs
```diff
@@ -1909,61 +1909,114 @@ node_id_property_methods! {
...
+    /// ARIA equivalent: [accessible description].
+    ///
+    /// [accessible description]: https://www.w3.org/TR/wai-aria-1.2/#dfn-accessible-description
     (Description, description, set_description, clear_description),
+    /// The text content or string value of this node. ARIA equivalent:
+    /// [`aria-valuetext`] for range widgets.
+    ///
+    /// **Difference with ARIA:** Unlike `aria-valuetext`, this property is also
+    /// used for text controls and the text content of nodes such as [`Role::Label`].
```
DataTriny: Mentionning `Role::TextRun` as well seem critical as it's how editable and rich text is represented in AccessKit.

common/src/lib.rs
```diff
@@ -1909,61 +1909,114 @@ node_id_property_methods! {
...
+    ///
+    /// [`aria-valuetext`]: https://www.w3.org/TR/wai-aria-1.2/#aria-valuetext
     (Value, value, set_value, clear_value),
     /// A single character, usually part of this node's name, that can be pressed,
     /// possibly along with a platform-specific modifier, to perform
     /// this node's default action. For menu items, the access key is only active
     /// while the menu is active, in contrast with [`keyboard_shortcut`];
-    /// a single menu item may in fact have both properties.
+    /// a single menu item may in fact have both properties. ARIA has no distinct
```
DataTriny: I don't think there is much value in flagging when no ARIA equivalent exists. Furthermore linking to `aria-keyshortcuts` probably just add confusion.

common/src/lib.rs
```diff
@@ -1909,61 +1909,114 @@ node_id_property_methods! {
...
+    /// equivalent; [`aria-keyshortcuts`] corresponds to [`keyboard_shortcut`].
     ///
     /// [`keyboard_shortcut`]: Node::keyboard_shortcut
+    /// [`aria-keyshortcuts`]: https://www.w3.org/TR/wai-aria-1.2/#aria-keyshortcuts
     (AccessKey, access_key, set_access_key, clear_access_key),
     /// A way for application authors to identify this node for automated
-    /// testing purpose. The value must be unique among this node's siblings.
+    /// testing purposes. The value must be unique among this node's siblings.
+    /// There is no ARIA equivalent.
```
DataTriny: The web equivalent would be the DOM element ID, but again I don't think providing this information is useful.

common/src/lib.rs
```diff
@@ -1909,61 +1909,114 @@ node_id_property_methods! {
...
     /// to support third-party math accessibility products that parse MathML.
+    /// There is no ARIA equivalent.
     (InnerHtml, inner_html, set_inner_html, clear_inner_html),
     /// A keystroke or sequence of keystrokes, complete with any required
     /// modifiers(s), that will perform this node's default action.
-    /// The value of this property should be in a human-friendly format.
+    /// The value of this property should be in a human-friendly format. ARIA
+    /// equivalent: [`aria-keyshortcuts`].
+    ///
+    /// **Difference with ARIA:** `aria-keyshortcuts` uses a standardized syntax,
```
DataTriny: Pretty sure we don't want human-friendly format here, but we'll revisit this later.

common/src/lib.rs
```diff
@@ -1909,61 +1909,114 @@ node_id_property_methods! {
...
+    /// while AccessKit expects a human-friendly string.
+    ///
+    /// [`aria-keyshortcuts`]: https://www.w3.org/TR/wai-aria-1.2/#aria-keyshortcuts
     (KeyboardShortcut, keyboard_shortcut, set_keyboard_shortcut, clear_keyboard_shortcut),
     /// An [IETF language tag](https://www.rfc-editor.org/info/bcp47).
-    /// Only present when different from parent.
+    /// Only set this when it differs from the parent. There is no ARIA equivalent;
+    /// web content supplies this information through the host language.
```
DataTriny: I don't think we want to talk about web content specifically. Saying that the default value is dictated by the host platform is obvious but maybe needs to be documented? @PoignardAzur

common/src/lib.rs
```diff
@@ -1909,61 +1909,114 @@ node_id_property_methods! {
...
+    /// [`aria-roledescription`]: https://www.w3.org/TR/wai-aria-1.2/#aria-roledescription
     (RoleDescription, role_description, set_role_description, clear_role_description),
     /// An optional string that may override an assistive technology's
     /// description of the node's state, replacing default strings such as
     /// "checked" or "selected". Note that most platform accessibility APIs
-    /// and assistive technologies do not support this feature.
+    /// and assistive technologies do not support this feature. There is no ARIA
+    /// equivalent.
     (StateDescription, state_description, set_state_description, clear_state_description),
     /// If a node's only accessible name comes from a tooltip, it should be
```
DataTriny: Could you reword this so we only mention node label and not accessible name?

common/src/lib.rs
```diff
@@ -1909,61 +1909,114 @@ node_id_property_methods! {
...
-    /// exposed through this property rather than [`label`].
+    /// exposed through this property rather than [`label`]. There is no direct
+    /// ARIA equivalent.
     ///
     /// [`label`]: Node::label
     (Tooltip, tooltip, set_tooltip, clear_tooltip),
+    /// The target URL of a link. There is no ARIA equivalent; web content
```
DataTriny: Talking about "host language" here doesn't make much sense, copy/paste mistake maybe? Anyway in the context of AccessKit we are the "host".

common/src/lib.rs
```diff
@@ -1909,61 +1909,114 @@ node_id_property_methods! {
...
+    /// equivalent: [`aria-rowindextext`].
+    ///
+    /// [`aria-rowindextext`]: https://www.w3.org/TR/wai-aria-1.3/#aria-rowindextext
     (RowIndexText, row_index_text, set_row_index_text, clear_row_index_text),
+    /// A human-readable alternative to the numeric column index. ARIA 1.3 draft
+    /// equivalent: [`aria-colindextext`].
+    ///
+    /// [`aria-colindextext`]: https://www.w3.org/TR/wai-aria-1.3/#aria-colindextext
     (ColumnIndexText, column_index_text, set_column_index_text, clear_column_index_text),
+    /// A concise label for this node in Braille. ARIA 1.3 draft equivalent:
```
DataTriny: Documentation of this property and the next one are confusing. One might think you have to write using the Unicode characters representing Braille dots here. In reality these are alternatives to `Node::label` and `Node::role_description` that assistive technologies can communicate to Braille displays, as opposed to speech synthesizers for instance.

DataTriny (changes requested):
Thank you for starting this.

Overall I think the format is good, except mentionning when there is no equivalent.

common/src/lib.rs
blinding-pixels: Hello, happy to get this rolling. I have made the requested changes. I can see why you wouldn't want the keyboard shortcut to be in Human friendly language, however since  there is nothing explicitly titled Screen Reader hint, maybe we move that  to the node description? I think for  people who might not be very well versed in screen readers knowing where to put that could be helpful. But its also a product choice , so I'll defer to your judgement for that.

common/src/lib.rs
```diff
@@ -1909,65 +1909,57 @@ node_id_property_methods! {
...
-    /// while AccessKit expects a human-friendly string.
-    ///
     /// [`aria-keyshortcuts`]: https://www.w3.org/TR/wai-aria-1.2/#aria-keyshortcuts
     (KeyboardShortcut, keyboard_shortcut, set_keyboard_shortcut, clear_keyboard_shortcut),
     /// An [IETF language tag](https://www.rfc-editor.org/info/bcp47).
-    /// Only set this when it differs from the parent. There is no ARIA equivalent;
-    /// web content supplies this information through the host language.
+    /// Only set this when it differs from the parent.
     (Language, language, set_language, clear_language),
     /// If a text input has placeholder text, it should be exposed
```
DataTriny: I think we're missing an introductory sentence to explain the purpose.

From ARIA: "A short hint (a word or short phrase) intended to aid the user with data entry when the control has no value. A hint could be a sample value or a brief description of the expected format."

Then reword the mention of `Node::label` as ARIA does too.

common/src/lib.rs
DataTriny, line 1972: Why have you removed the ARIA equivalent here? We probably also want to rephrase the introductory sentence to explain why it can be useful. Not as long as ARIA though.

DataTriny (commented):
It's looking good. Just two more improvements from me.

I'd value @PoignardAzur's opinion on this before merging.

common/src/lib.rs
blinding-pixels: The role description still has its ARIA equivalent documented. Did you mean `StateDescription`? I removed the “There is no ARIA equivalent” sentence from that property based on your earlier request not to mention properties without ARIA equivalents.

common/src/lib.rs
DataTriny: We don't expose this property on any platform yet, we will make changes when we actually implement it.

common/src/lib.rs
DataTriny: I was indeed looking at the wrong diff, sorry.

common/src/lib.rs
DataTriny: Not sure if we clearly communicate the fact that placeholder and label can be used in conjunction, they serve different purposes. ARIA does it well, at least people can click on the link.

DataTriny (approved):
Looks good to me.

Let's give Olivier a chance to go through this.

PoignardAzur (approved):
I tried running `cargo doc` and the generated docs seem fine. I think the format is good as-is.

Some of the descriptions seem a little vague about how this data will be interpreted, but that's par for the course with accessibility standards, and this will do for now.

Thanks for the PR!

blinding-pixels:
Will do. And you're most welcome. Happy to be of help. Please let me know the next set of properties you would like mapped, or if you like I can go in and start going through them one by one and do this here on out.

PoignardAzur:
> or if you like I can go in and start going through them one by one and do this here on out.

Sounds good.

PR #756, opened 2026-07-30
fix: Label via accessibilityValue for some roles on macOS

DataTriny:
See https://github.com/slint-ui/slint/issues/12644

The logic comes straight from Chromium.

### DIFF
--- a/adapters/macos/src/node.rs
+++ b/adapters/macos/src/node.rs
@@ -230,6 +230,17 @@ fn ns_role(node: &NodeRef) -> &'static NSAccessibilityRole {
     }
 }
 
+fn label_is_exposed_in_value(role: Role) -> bool {
+    matches!(
+        role,
+        Role::Label
+            | Role::ListBoxOption
+            | Role::ListMarker
+            | Role::MenuListOption
+            | Role::TitleBar
+    )
+}
+
 fn ns_sub_role(node: &NodeRef) -> &'static NSAccessibilitySubrole {
     let role = node.role();
 
@@ -309,6 +320,9 @@ impl NodeWrapper<'_> {
             // includes a title, VoiceOver behavior is broken.
             return None;
         }
+        if label_is_exposed_in_value(self.0.role()) {
+            return None;
+        }
         self.0.label()
     }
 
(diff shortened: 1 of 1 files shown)


DataTriny:
@John-Howard Would you please be able to check if this fixes the issue you reported? I currently don't have access to my Mac.

I tested Slint's ListView on macOS 11 a while back and VoiceOver was correctly reporting item names, the behavior may have changed in later macOS versions?

Could this by chance fix #520 at the same time?

Thanks in advance for your help.

John-Howard:
Tested on macOS 26.6, Slint 1.17.1 (winit backend), with this branch patched in via `[patch.crates-io]` on `accesskit` / `accesskit_consumer` / `accesskit_macos`. App under test is a real 12-row Slint `ListView` whose rows are `accessible-role: list-item` + `accessible-label`.

**The tree change works as intended.** Before: row is `AXStaticText`, `AXTitle="Synthetic project 0"`, `AXValue` missing. After: `AXTitle` missing, `AXValue="Synthetic project 0"`. Window title and `Role::Button` toolbar controls are byte-identical across both builds.

**But the rows were never silent on 26.6, so there's no symptom here for it to fix.** Captured via VoiceOver's own AppleScript (`content of last phrase`) while arrowing through rows — stock: `Synthetic task 3 text`; this branch: `Synthetic task 3`. VoiceOver already reads the label out of `AXTitle` for `ListBoxOption`; the PR's audible effect is to drop the trailing `text` role description. **No double-announcement in either build**, listening by ear.

**Our [slint#12644](https://github.com/slint-ui/slint/issues/12644) report was wrong about the cause, and I'm sorry for sending you after it.** The rows were silent because AX focus never landed on a row — `accessible-delegate-focus` resolving against materialised children, which we filed separately as [slint#12645](https://github.com/slint-ui/slint/issues/12645). Both fixes shipped in one commit and I credited the wrong one. Isolated now by building the pre-fix commit: no row ever reports `AXFocused`, and VoiceOver stays on a toolbar button. Two further corrections: the version field said macOS 15, but it was macOS 26.5.1; and the reproducer as filed isn't representative — its child `Text` has no `accessible-role: none`, so it's a `Role::Label` node carrying the title in its own `AXValue`, and pasting that snippet does not produce an unreadable row.

**On #520:** unaffected either way. `AXSelected` is exposed and correct on both builds (Slint calls `set_selected`) — exactly one row true — yet VoiceOver announced no selection state in either. So #520 looks like an announcement issue rather than a missing attribute, and this PR neither fixes nor regresses it.

No objection to the change — it's harmless here and slightly terser — but I can't report it as fixing a real symptom on current macOS.

mwcampbell (approved):
Looks good to me.

PR #757, opened 2026-07-30
Document ARIA mappings for usize properties

blinding-pixels:
This continues the documentation work from #753 by covering the `usize` properties discussed in #749, using the same format established in #753.

A couple of points I would particularly appreciate confirmation on:

- AccessKit uses zero-based indexes, whereas the corresponding ARIA properties use one-based indexes. This also led me to correct the `PositionInSet` documentation: its value must be less than `SizeOfSet`, rather than less than or equal to it.
- AccessKit places `SizeOfSet` on the container, whereas ARIA places `aria-setsize` on each item.

I ran:

- `cargo test -p accesskit`

### DIFF
--- a/accesskit/src/lib.rs
+++ b/accesskit/src/lib.rs
@@ -2035,19 +2035,67 @@ f32_property_methods! {
 }
 
 usize_property_methods! {
+    /// The total number of rows in a table, grid, or tree grid. ARIA equivalent:
+    /// [`aria-rowcount`].
+    ///
+    /// [`aria-rowcount`]: https://www.w3.org/TR/wai-aria-1.2/#aria-rowcount
     (RowCount, row_count, set_row_count, clear_row_count),
+    /// The total number of columns in a table, grid, or tree grid. ARIA
+    /// equivalent: [`aria-colcount`].
+    ///
+    /// [`aria-colcount`]: https://www.w3.org/TR/wai-aria-1.2/#aria-colcount
     (ColumnCount, column_count, set_column_count, clear_column_count),
+    /// The row index of this node within a table, grid, or tree grid. ARIA
+    /// equivalent: [`aria-rowindex`].
+    ///
+    /// **Difference with ARIA:** `aria-rowindex` is one-based, while this
+    /// AccessKit property is zero-based.
+    ///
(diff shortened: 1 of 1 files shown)


accesskit/src/lib.rs
```diff
@@ -2035,19 +2035,67 @@ f32_property_methods! {
...
+    /// The total number of columns in a table, grid, or tree grid. ARIA
+    /// equivalent: [`aria-colcount`].
+    ///
+    /// [`aria-colcount`]: https://www.w3.org/TR/wai-aria-1.2/#aria-colcount
     (ColumnCount, column_count, set_column_count, clear_column_count),
+    /// The row index of this node within a table, grid, or tree grid. ARIA
+    /// equivalent: [`aria-rowindex`].
+    ///
+    /// **Difference with ARIA:** `aria-rowindex` is one-based, while this
+    /// AccessKit property is zero-based.
```
DataTriny: I think the word AccessKit here is unnecessary. Can you remove it to the other zero-based properties below?

DataTriny (changes requested):
Just one comment.

accesskit/src/lib.rs
```diff
@@ -2035,19 +2035,67 @@ f32_property_methods! {
...
+    /// **Difference with ARIA:** `aria-level` is one-based, while this property
+    /// is zero-based.
+    ///
+    /// [`aria-level`]: https://www.w3.org/TR/wai-aria-1.2/#aria-level
     (Level, level, set_level, clear_level),
     /// For containers like [`Role::ListBox`], specifies the total number of items.
+    /// ARIA equivalent: [`aria-setsize`].
+    ///
+    /// **Difference with ARIA:** `aria-setsize` is set on each item, while this
+    /// AccessKit property is set on the container.
```
DataTriny:
```suggestion
    /// property is set on the container.
```

DataTriny (changes requested):
Forgot about another one.

DataTriny (approved):
Thanks again for this one.

blinding-pixels:
My Pleasure. Working on the next one already should have that iopened in a bit.

PR #760, opened 2026-08-05
Document ARIA mappings for boolean and node-list properties

blinding-pixels:
This continues the documentation work from #753 and #757 by covering two boolean properties and eight node-list properties discussed in #749.

Two of these properties, `Children` and `RadioGroup`, do not have direct ARIA equivalents. One property, `Details`, differs from its ARIA 1.2 equivalent because `aria-details` references a single element, while `Details` can reference multiple nodes. This difference has been documented.

I ran:

- `cargo test -p accesskit`

Happy to revise anything if required.

### DIFF
--- a/accesskit/src/lib.rs
+++ b/accesskit/src/lib.rs
@@ -1878,20 +1878,42 @@ vec_type_methods! {
 }
 
 node_id_vec_property_methods! {
+    /// The ordered list of this node's direct children in the tree.
     (Children, children, set_children, push_child, clear_children),
+    /// The nodes whose contents or presence are controlled by this node. ARIA
+    /// equivalent: [`aria-controls`].
+    ///
+    /// [`aria-controls`]: https://www.w3.org/TR/wai-aria-1.2/#aria-controls
     (Controls, controls, set_controls, push_controlled, clear_controls),
+    /// The nodes that provide detailed, extended descriptions for this node.
+    /// ARIA equivalent: [`aria-details`].
+    ///
+    /// **Difference with ARIA:** `aria-details` references a single element,
+    /// while this property may reference multiple nodes.
+    ///
+    /// [`aria-details`]: https://www.w3.org/TR/wai-aria-1.2/#aria-details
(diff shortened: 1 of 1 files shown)


accesskit/src/lib.rs
```diff
@@ -1878,20 +1878,42 @@ vec_type_methods! {
...
 
 node_id_vec_property_methods! {
+    /// The ordered list of this node's direct children in the tree.
     (Children, children, set_children, push_child, clear_children),
+    /// The nodes whose contents or presence are controlled by this node. ARIA
+    /// equivalent: [`aria-controls`].
+    ///
+    /// [`aria-controls`]: https://www.w3.org/TR/wai-aria-1.2/#aria-controls
     (Controls, controls, set_controls, push_controlled, clear_controls),
+    /// The nodes that provide detailed, extended descriptions for this node.
```
DataTriny: I feel like this documentation is too similar to the one for described_by. ARIA explains the link between the two and maybe we should do the same.

accesskit/src/lib.rs
```diff
@@ -1878,20 +1878,42 @@ vec_type_methods! {
 }
 
 node_id_vec_property_methods! {
+    /// The ordered list of this node's direct children in the tree.
```
DataTriny: I don't know if the single word "ordered" here is enough. Maybe we should replace it by a second sentence to hammer the fact that the order must be chosen carefully as it directly influences how widgets are placed, akin to the tab index on a web page. Any idea?

DataTriny (changes requested):

accesskit/src/lib.rs
blinding-pixels: Wanted to check, are we trying to leave this as documentation alone, or are we  also maybe trying to guide people who might be  new to accessibility on the best practices?

DataTriny: I think your latest changes are enough for users to understand the importance of the order of this list.

DataTriny (approved):
Thanks once more for this one.

PR #762, opened 2026-08-09
Expose EditableText.SetTextContents on Unix

### LINKED ISSUE #761: Expose AT-SPI EditableText.SetTextContents as Action::SetValue
`accesskit_unix` exposes the AT-SPI `Text` interface, but not `EditableText`. AT-SPI clients can read and focus text inputs, but they cannot replace the contents through the standard editable-text API.

Would it make sense to start with `EditableText.SetTextContents` and forward it as an `ActionRequest` using `Action::SetValue` and `ActionData::Value`? Methods without matching AccessKit actions could remain unsupported.

PR #362 intentionally left this interface out because Orca did not use it and Chromium did not implement it. It is useful for desktop automation clients, though. I ran into this with `computer-use-mcp` against egui and Bevy applications on Wayland: numeric controls expose AT-SPI `Value`, while text inputs have no equivalent string-setting capability.

I am happy to prototype the minimal mapping.

mirsella:
Closes #761.

This exposes AT-SPI `EditableText` for writable text inputs that handle string `SetValue` requests. `SetTextContents` forwards the new contents as `Action::SetValue` with `ActionData::Value`.

Numeric controls continue to use AT-SPI `Value`. The remaining `EditableText` methods return `NotSupported` because AccessKit does not have matching clipboard or range-editing actions.


adapters/atspi-common/src/node.rs
```diff
@@ -438,6 +437,15 @@ impl NodeWrapper<'_> {
         self.0.raw_bounds().is_some() || self.is_root()
     }
 
+    fn supports_editable_text(&self) -> bool {
+        // Empty inputs may have no text ranges. Numeric controls use AT-SPI Value because
```
DataTriny: This comment is just wrong, please remove it.

adapters/atspi-common/src/node.rs
```diff
@@ -438,6 +437,15 @@ impl NodeWrapper<'_> {
...
+        // SetValue doesn't declare which ActionData variant the handler accepts.
+        self.0.is_text_input()
```
DataTriny: For consistency with other adapters the condition here should just be: `self.is_text_input() && self.supports_text_ranges()`

adapters/atspi-common/src/node.rs
```diff
@@ -759,20 +770,32 @@ impl PlatformNode {
         self.resolve_for_text_with_context(|node, _, _| f(node))
     }
 
-    fn do_action_internal<F>(&self, target: FullNodeId, f: F) -> Result<()>
-    where
-        F: FnOnce(&TreeState, &Context, NodeId, TreeId) -> ActionRequest,
-    {
+    fn dispatch_action(&self, action: Action, data: Option<ActionData>) -> Result<()> {
```
DataTriny: We already know that a node implements the interface because the D-Bus object exists, no need to check it again here. You could check for the read-only property in `PlatformNode::set_text_contents` but I'd get rid of this entire refactoring. Don't worry too much about the simplified API.

adapters/unix/src/atspi/interfaces/editable_text.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    fn paste_text(&self, _position: i32) -> fdo::Result<bool> {
+        Err(unsupported())
+    }
+
+    fn set_text_contents(&self, new_contents: &str) -> fdo::Result<bool> {
+        self.0
+            .set_text_contents(new_contents)
+            .map_err(|error| crate::util::map_error_from_node(&self.0, error))
```
DataTriny: The other interfaces define a `map_error` method, please do the same.

adapters/atspi-common/src/adapter.rs
```diff
@@ -691,15 +695,69 @@ mod tests {
         }
     }
 
+    #[test]
+    fn editable_text_support_and_dispatch() {
```
DataTriny: These tests only cover the behavior of nodes so they should probably live in the node module. We may need to move some shared utilities in `lib.rs`.

DataTriny (changes requested):
Hi @mirsella, thanks for that. A couple of comments below.

mirsella:
hello ! thanks for the review :)
ive addressed your comments.
ive forget to mention, this PR is mostly AI generated. feel free to close this if this doesn't align with you !

DataTriny:
@mirsella I kindly ask you to not send such contribution again. I spent more time spotting issues in your changes than I would have writing the feature myself.

You have been warned.

mirsella:
i am very sorry to have wasted your time. sorry again to not have specified it sooner.
thanks you for all your work !

PS:
if someones wants do re use my fork, feel free to do whatever of course.
im using it for my project, and ill soon implement web accessibility support too

PR #763, opened 2026-08-12
Reuse property buffers when cloning nodes

### LINKED ISSUE #444: Should accesskit use `Box<str>`?

DataTriny:
Closes #444 

Extracted from the [stacked-alloc-refactors](https://github.com/AccessKit/accesskit/tree/stacked-alloc-refactors) branch.

We can save a lot of memory allocations just with this when we move nodes around.

### DIFF
--- a/accesskit/src/lib.rs
+++ b/accesskit/src/lib.rs
@@ -14,7 +14,7 @@ extern crate alloc;
 
 #[cfg(feature = "schemars")]
 use alloc::borrow::Cow;
-use alloc::{boxed::Box, string::String, vec::Vec};
+use alloc::{borrow::ToOwned, boxed::Box, string::String, vec::Vec};
 use core::fmt;
 #[cfg(feature = "pyo3")]
 use pyo3::pyclass;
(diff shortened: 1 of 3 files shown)


accesskit/src/lib.rs
```diff
@@ -1084,6 +1133,20 @@ impl Properties {
...
+            values: self.values.clone(),
+        }
+    }
+
+    fn clone_from(&mut self, source: &Self) {
+        self.indices = source.indices;
+        source.values.clone_into(&mut self.values);
+    }
+}
+
```
PoignardAzur, line 1149: I know it's a little late to comment, but I was wondering what's the rationale for this manual clone impl? It seems to me like the derive would do the exact same thing?

DataTriny: We need to manually implement Clone for Properties so we can reuse the value buffers in `clone_from`. The derive impl doesn't do that. It save us allocations when we move nodes between tree states.

PoignardAzur: > The derive impl doesn't do that.

Oh wow, I really didn't expect that. Thanks for the explanation.

PR #764, opened 2026-08-14
Document ARIA mappings for enum-valued properties

blinding-pixels:
This continues the documentation work by covering the enum-valued properties.

The two differences I found are that `Toggled` combines the separate `aria-checked` and `aria-pressed` states based on the node's role, and `AriaCurrent` supports the defined enum values while ARIA also treats unrecognized non-empty tokens as `true`.

I ran:

- `cargo test -p accesskit`

### DIFF
--- a/accesskit/src/lib.rs
+++ b/accesskit/src/lib.rs
@@ -2294,10 +2294,6 @@ unique_enum_property_methods! {
     /// Whether and how this node represents the current item within a set. ARIA
     /// equivalent: [`aria-current`].
     ///
-    /// **Difference with ARIA:** `aria-current` accepts unrecognized non-empty
-    /// tokens and treats them as `true`, while this property accepts only the
-    /// values defined by [`AriaCurrent`].
-    ///
     /// [`aria-current`]: https://www.w3.org/TR/wai-aria-1.2/#aria-current
     (AriaCurrent, aria_current, set_aria_current, clear_aria_current, True),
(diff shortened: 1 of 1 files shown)


accesskit/src/lib.rs
```diff
@@ -2259,18 +2259,62 @@ bool_property_methods! {
...
     (Orientation, orientation, set_orientation, clear_orientation, Vertical),
+    /// The direction in which items are sorted by this row or column header.
+    /// ARIA equivalent: [`aria-sort`].
+    ///
+    /// [`aria-sort`]: https://www.w3.org/TR/wai-aria-1.2/#aria-sort
     (SortDirection, sort_direction, set_sort_direction, clear_sort_direction, Descending),
+    /// Whether and how this node represents the current item within a set. ARIA
+    /// equivalent: [`aria-current`].
+    ///
+    /// **Difference with ARIA:** `aria-current` accepts unrecognized non-empty
```
DataTriny: The difference is obvious and already enforced by the type system, I don't think it's even worth mentioning here.

DataTriny (changes requested):
Looks good, just one comment.

accesskit/src/lib.rs
blinding-pixels: Done. Apologies for the deleay.

DataTriny (approved):
Thanks!

PR #770, opened 2026-08-21
Update phf to version 0.14

SimonSapin:
This will help Servo avoid having two copies of phf when updating other dependencies to versions that use phf 0.14


DataTriny (approved):
Thank you. Expect a new release in a few days.

PR #772, opened 2026-08-23
docs: Document node flag properties

blinding-pixels:
This continues the documentation work. It covers the flag properties.

The main difference I found is that `Hidden` excludes a node from hit testing as well as from the tree presented to assistive technologies, while `aria-hidden` only controls exposure to assistive technologies.

The spelling and grammar error flags are documented separately from `aria-invalid`, since they identify marked text rather than the invalid state of an input.

I ran:

- `cargo test -p accesskit`

### DIFF
--- a/accesskit/src/lib.rs
+++ b/accesskit/src/lib.rs
@@ -1879,36 +1879,70 @@ impl Node {
 }
 
 flag_methods! {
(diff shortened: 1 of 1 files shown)


DataTriny (approved):
Thanks Eshaan!

blinding-pixels:
Anytime :)

PR #776, opened 2026-08-24
feat: Add html_id node property

### LINKED ISSUE #705: Add a property to allow exposing DOM ID
We're working on the [Servo accessibility](https://github.com/servo/servo/issues/4344) implementation using accesskit, and we've run into a stumbling block where we seem to have no way to reliably expose the [`id` attribute](https://www.w3.org/TR/html-aam-1.0/#att-id) to platform APIs.

The closest seems to be [`author_id`](https://docs.rs/accesskit/0.22.0/accesskit/struct.Node.html#method.author_id) which is mapped to `UIA_AutomationIdPropertyId` in UIA and `id` in Atspi, but isn't mapped to `AXDOMIdentifier` in AX API. It may not even be appropriate to map it through to that property.

We're interested to begin using the [WPT for AAMs](https://github.com/web-platform-tests/rfcs/blob/main/rfcs/wpt-for-aams.md) infrastructure, which requires exposing the DOM ID as a way to find an accessibility node to make assertions about.

DataTriny:
Closes #705

### DIFF
--- a/accesskit/src/lib.rs
+++ b/accesskit/src/lib.rs
@@ -918,6 +918,7 @@ enum PropertyId {
     AuthorId,
     ClassName,
     FontFamily,
+    HtmlId,
     HtmlTag,
     InnerHtml,
     KeyboardShortcut,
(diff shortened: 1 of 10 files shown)


DataTriny:
@alice this would be enough to unblock your WPT work, right?

alice:
Yes, thank you so much!

alice:
I will add, though - I spoke with the author of those tests, and she suggested we could also have the tests look for _both_ `AXDOMIdentifier` and `AXIdentifier`, so we might also be ok without this as far as the tests go.

Still, it will probably be useful since that is still what the HTML-AAM mappings specify at least for now.

PR #781, opened 2026-08-28
docs: Document remaining node properties

blinding-pixels:
Hey, this covers the 23 remaining properties that did not already have complete documentation. There are 12 more that already had documentation; I could do a consistency pass on those 12 if you like. Happy to tackle whatever else next.

Ran:

- `cargo test -p accesskit`


DataTriny (approved):

DataTriny:
I'm happy with the style you have applied in your changes. If you think existing documentation is now inconsistent, feel free to submit patches for that. Otherwise we have a bunch of types that currently don't have documentation in the accesskit crate, if you want to continue this type of work. Some of these types also have an ARIA equivalent. Except for the `Role` enum for which we need to find a good format, everything else is up for grab.

PR #790, opened 2026-09-04
feat(unix): implement AT-SPI Table and TableCell interfaces

rit3sh-x:
## Summary

- Implements `org.a11y.atspi.Table` and `org.a11y.atspi.TableCell` on the Unix (AT-SPI) adapter - the two interfaces `accesskit_unix` was missing for table/grid content.
- Table geometry is derived live from `Role::Row` children, falling back to `row_index`/`column_index` for flat ARIA-grid-style tables that have no `Row` children.
- All data already existed on `accesskit::Node` - this is AT-SPI glue, not new data modeling.
- Fixes a design bug found via testing: row/column selection queries (`IsRowSelected`, `GetSelectedColumns`, ...) now only count `Cell`/`GridCell` children, excluding `RowHeader`/`ColumnHeader`. Apps commonly never call `set_selected` on header cells, so counting them would permanently block any headered table from ever reporting a selected row/column.

## Test plan

- [x] `cargo test` (workspace default members)
- [x] `cargo test -p accesskit_atspi_common` (8 new unit tests covering interface gating, row/col counts, `GetIndexAt`/`GetRowAtIndex`/`GetColumnAtIndex` round-trip, headers/descriptions, selection queries, `AddRowSelection` action dispatch)
- [x] `cargo test -p accesskit_unix` (3 new tests exercising the D-Bus-facing `TableInterface` directly)
- [x] `cargo clippy -p accesskit_atspi_common -p accesskit_unix --all-targets -- -D warnings`
- [x] `cargo fmt --check`

### DIFF
--- a/adapters/atspi-common/src/error.rs
+++ b/adapters/atspi-common/src/error.rs
@@ -5,7 +5,7 @@
 
 use std::fmt;
 
-#[derive(Debug)]
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub enum Error {
     Defunct,
     UnsupportedInterface,
--- a/adapters/atspi-common/src/node.rs
+++ b/adapters/atspi-common/src/node.rs
@@ -485,6 +485,14 @@ impl NodeWrapper<'_> {
         self.0.is_container_with_selectable_children()
     }
 
+    fn supports_table(&self) -> bool {
+        matches!(self.0.role(), Role::Table | Role::Grid | Role::ListGrid)
+    }
+
+    fn supports_table_cell(&self) -> bool {
+        matches!(self.0.role(), Role::Cell | Role::GridCell)
+    }
+
     fn supports_text(&self) -> bool {
         self.0.supports_text_ranges()
     }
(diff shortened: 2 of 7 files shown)


rit3sh-x:
@DataTriny could you review it?

DataTriny:
Hello @rit3sh-x,

Did you use an LLM to write all of this?

rit3sh-x:
@DataTriny I did use it for writing some tests. Is there some AI Policy?

rit3sh-x:
I was also working on the Image and HyperText APIs.

DataTriny:
@rit3sh-x I only reviewed a small portion of your changes, which is enough for me to conclude that you either used an LLM way more than you claim, or you didn't really understand what you were doing. I am not going to review the rest.

Quick note on the tui-a11y project you mention in the PR description: proper terminal emulators already emit their own accessibility tree with the content of the text buffer, any scroll bar, tabs... And they know when their window gains or loses focus. While the AT-SPI protocol doesn't enforce it, you must not create an accessibility tree if the application already emits one, this responsibility falls on who creates and manages the window. Your library is only going to confuse assistive technologies like Orca.

Please stop that.

Respectfully.

rit3sh-x:
Sorry if it felt like the code was fully LLM generated. I did use an LLM for some of the tests, but I worked through the implementation myself.

My main motivation for tui-a11y was accessibility for lower-level TUIs/CLIs, rather than advanced terminal emulators like Warp that already have their own accessibility tree.

I appreciate the feedback, and I'll try to make sure my code doesn't come across as AI-generated in the future.

